PR #22, opened 2021-05-02
gh-14 Add support for nearImage argument

antas-marcin:


v2/weaviate/graphql/nearimagebuilder.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+// WithImage base64 encoded image
+func (b *NearImageArgumentBuilder) WithImage(image string) *NearImageArgumentBuilder {
+	b.image = image
+	return b
+}
+
+// WithFileReader the image file
+func (b *NearImageArgumentBuilder) WithFileReader(fileReader io.Reader) *NearImageArgumentBuilder {
```
etiennedi, line 25: I think we can give this one a more generic name, since would also work different readers, such as `bytes.Reader` or basically anything with a `Read([]byte) (int,error)` method.

How about just `.WithReader()`?

antas-marcin: sounds good!

PR #37, opened 2021-12-13
Feature: Create data object with custom vectors

### LINKED ISSUE #36: Creating a data object with custom vectors
Hello!

I noticed in the [Weaviate API docs](https://www.semi.technology/developers/weaviate/current/restful-api-references/objects.html#create-a-data-object-with-custom-vectors) that "`creating a data object with custom vectors will be supporting in the GO client soon`", however I don't see an open issue or PR that address this, and it is a piece of functionality that would be very useful to me.

Do you mind if I implement data object creation with custom vectors, and open an PR?

Thanks!

parkerduckworth:
## Summary

- Implement `WithVector` method on the `data.Creator` type.
- Write integration tests asserting correctness
- Fix typo found in comment

---

Closes #36

### DIFF
--- a/v2/test/data/data_test.go
+++ b/v2/test/data/data_test.go
@@ -35,6 +35,10 @@ func TestData_integration(t *testing.T) {
 			"name":        "ChickenSoup",
 			"description": "Used by humans when their inferior genetics are attacked by microscopic organisms.",
 		}
+		propertySchemaV := map[string]string{
+			"name":        "Supreme",
+			"description": "Contains all toppings known to humankind.",
+		}
 
 		wrapperT, errCreateT := client.Data().Creator().
 			WithClassName("Pizza").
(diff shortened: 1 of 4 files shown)


parkerduckworth:
I agree with the CLA.

etiennedi:
Thanks @parkerduckworth, 

before we can merge, there are two more things that need to be added:

1. Your current `WithVector()` method only stores the vector in the `Creator`, but does not actually send it off to Weaviate. For this you also need to add it to method which builds the `PayloadObject()`: https://github.com/semi-technologies/weaviate-go-client/blob/c92ea0f3b4356e68191a3e7b73a22f378a58a04d/v2/weaviate/data/creator.go#L70-L79
2. The current test does not test if the custom vector is set correctly. Otherwise, I think it would have caught the above. To do so, we need to create a class that has no vectorizer set. (The existing tests assume everything is vectorized using the `text2vec-contextionary` module). Here's an example of the same process in the Java client where the tests creates an explicit class with `vectorizer: none` to verify that the custom vector is set correctly: https://github.com/semi-technologies/weaviate-java-client/pull/73/files#diff-7d48439c8b3130dea97f466b3e3769b002b12db9e0c53018ac7002a29b858c48R642-R691

Let me know how you want to proceed. Would you like to make the additional changes or should we?

Thanks
Etienne

parkerduckworth:
@etiennedi thanks for the pointers, I'll push up the changes today!

etiennedi:
Hi @parkerduckworth, awesome, thanks for the progress. I've pushed one more tiny addition: https://github.com/semi-technologies/weaviate-go-client/pull/37/commits/4f0114429d32216bdb9d7f8cd4cc6c27b5acf1c5

While your test was already setting up everything correctly, it didn't yet assert that the custom vector was indeed set. I've added a simple assertion to make sure that the retrieved vector matches the one at creation.

Once CI is green, I'll merge and release. Thank you very much for your contribution!

parkerduckworth:
Ah yes, good catch.  Thanks @etiennedi !

PR #52, opened 2022-03-31
gh-30 Add group filter builder

### LINKED ISSUE #30: Add group filter builder

antas-marcin:
Closes #30

### DIFF
--- a/v3/weaviate/graphql/groupbuilder.go
+++ b/v3/weaviate/graphql/groupbuilder.go
@@ -1,0 +1,46 @@
+package graphql
+
+import (
+	"fmt"
+	"strings"
+)
+
+// GroupType filter
+type GroupType string
+
+// Merge group type filter
+const Merge GroupType = "merge"
+
+// Closest group type filter
+const Closest GroupType = "closest"
+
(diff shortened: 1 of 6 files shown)


v3/weaviate/graphql/groupbuilder.go
```diff
@@ -0,0 +1,46 @@
...
+// Build build the given clause
+func (b *GroupArgumentBuilder) build() string {
+	clause := []string{}
+	if len(b.withType) > 0 {
+		clause = append(clause, fmt.Sprintf("type: %s", b.withType))
+	}
+	if b.withForce {
+		clause = append(clause, fmt.Sprintf("force: %v", b.force))
+	}
+	return fmt.Sprintf("group:{%s}", strings.Join(clause, " "))
```
parkerduckworth, line 45: Does this still work when `clause` is empty, assuming the result in that case would be `group{}`?

antas-marcin: it will fail when it's not empty

parkerduckworth (approved):
LGTM

PR #57, opened 2022-04-01
travis scripts improved to be zsh compatible

aliszka:


parkerduckworth (approved):
Looks good. I'm surprised this is valid bash in the first place

PR #65, opened 2022-04-19
Implement nearMedia filters for Aggregate query

parkerduckworth:


v4/test/graphql/graphql_test_helper.go
```diff
@@ -0,0 +1,53 @@
+package graphql
```
antas-marcin, line 1: shouldn't the file be named `graphql_helper_test.go`? are we planning to use this file outside tests?

parkerduckworth: fixed

v4/weaviate/graphql/aggregatequerybuilder.go
```diff
@@ -0,0 +1,117 @@
+package graphql
```
antas-marcin, line 1: I would leave the name `aggregate.go` I know that you wanted to align this change with `getquerybuilder.go` but for get we have `get.go` that uses `getquerybuilder.go` and that's why we have the name for this like `getquerybuilder.go` I think that we can leave the `aggregate.go`

v4/weaviate/graphql/nearvectorbuilder.go
```diff
@@ -0,0 +1,42 @@
...
+	return b
+}
+
+// Build build the given clause
+func (b *NearVectorArgumentBuilder) build() string {
+	clause := []string{}
+	if b.withCertainty {
+		clause = append(clause, fmt.Sprintf("certainty: %v", b.certainty))
+	}
+	if b.vector != nil {
```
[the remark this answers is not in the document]
parkerduckworth: done

antas-marcin (approved):

PR #77, opened 2022-05-19
TD-114: add support for mover objects as arguments to nearText

redouan-rhazouani:


v4/weaviate/graphql/neartextbuilder.go
```diff
@@ -6,6 +6,68 @@ import (
 	"strings"
 )
 
+// fldMover is a type representing field names of a move sub query
+type fldMover string
+
+const (
+	fldMoverConcepts fldMover = "concepts"
+	fLDMoverForce    fldMover = "force"
```
parkerduckworth, line 14: `fLDMoverForce` -> maybe fix to `fldMoverForce` to match naming convention of the other values

v4/weaviate/graphql/neartextbuilder.go
```diff
@@ -6,6 +6,68 @@ import (
...
+	// Concepts that should be used as base for the movement operation
+	Concepts []string
+	// Force to be applied in the movement operation
+	Force float32
+	// Movers _
+	Movers []MoverObject
+}
+
+func (m *MoveParameters) String() string {
+	concepts, _ := json.Marshal(m.Concepts)
```
parkerduckworth, line 31: are we certain this can never error? not sure what kind of validation is performed prior to this point

v4/weaviate/graphql/neartextbuilder.go
redouan-rhazouani: Typo of course 😁 will be fixed

v4/weaviate/graphql/neartextbuilder.go
redouan-rhazouani: This should never returns an error.

To be certain I've added a new utility which marshal the slice above without returning any error

v4/weaviate/graphql/neartextbuilder.go
```diff
@@ -6,6 +6,68 @@ import (
...
+	fldMoverObjects  fldMover = "objects"
+)
+
+// MoveParameters to fine tune Explore queries
+type MoveParameters struct {
+	// Concepts that should be used as base for the movement operation
+	Concepts []string
+	// Force to be applied in the movement operation
+	Force float32
+	// Movers _
```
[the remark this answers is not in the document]
redouan-rhazouani: thanks I've added a comment

v4/weaviate/graphql/neartextbuilder.go
```diff
@@ -1,11 +1,72 @@
...
+	}
+	if len(ms) < 1 {
+		return fmt.Sprintf("{%s: %s %s: %v}", fldMoverConcepts, concepts, fLdMoverForce, m.Force)
+	}
+
+	s := "{"
+	if len(m.Concepts) > 0 {
+		s = fmt.Sprintf("{%s: %s", fldMoverConcepts, concepts)
+	}
+	return fmt.Sprintf("%s %s: %v %s: %v}", s, fLdMoverForce, m.Force, fldMoverObjects, ms)
```
antas-marcin, line 45: fLdMoverForce here and in line 38 are not changed

v4/weaviate/graphql/neartextbuilder.go
```diff
@@ -1,11 +1,72 @@
...
+	// Concepts that should be used as base for the movement operation
+	Concepts []string
+	// Force to be applied in the movement operation
+	Force float32
+	// Movers is a list of movers used to adjust the serach direction
+	Movers []MoverObject
+}
+
+func (m *MoveParameters) String() string {
+	concepts := marshalStrings(m.Concepts)
```
antas-marcin, line 30: I would remove the `marshalStrings` method and make the code simpler and better to read by refactoring for example to this version:
```
clause := []string{}
if len(m.Concepts) > 0 {
    concepts := make([]string, len(m.Concepts))
    for i := range m.Concepts {
	concepts[i] = fmt.Sprintf("\"%s\"", m.Concepts[i])
   }
   clause = append(clause, fmt.Sprintf("concepts:[%s]", strings.Join(concepts, ", ")))
}
```

v4/weaviate/graphql/neartextbuilder.go
```diff
@@ -1,11 +1,72 @@
...
+)
+
+// MoveParameters to fine tune Explore queries
+type MoveParameters struct {
+	// Concepts that should be used as base for the movement operation
+	Concepts []string
+	// Force to be applied in the movement operation
+	Force float32
+	// Movers is a list of movers used to adjust the serach direction
+	Movers []MoverObject
```
antas-marcin, line 26: I would suggest to change the name of this field to reflect the actual GraphQL field `Movers` -> `Objects` so that it would be aligned with all other our clients and also users would know what's the purpose of this field as they would already relate to `nearText.moveTo.objects`

v4/weaviate/graphql/neartextbuilder.go
```diff
@@ -1,11 +1,72 @@
...
+			ms = append(ms, m.String())
+		}
+	}
+	if len(ms) < 1 {
+		return fmt.Sprintf("{%s: %s %s: %v}", fldMoverConcepts, concepts, fLdMoverForce, m.Force)
+	}
+
+	s := "{"
+	if len(m.Concepts) > 0 {
+		s = fmt.Sprintf("{%s: %s", fldMoverConcepts, concepts)
```
antas-marcin, line 43: I think that we are missing here a closing bracket?

v4/weaviate/graphql/neartextbuilder.go
```diff
@@ -1,11 +1,72 @@
...
+
+func (m *MoveParameters) String() string {
+	concepts := marshalStrings(m.Concepts)
+	ms := make([]string, 0, len(m.Movers))
+	for _, m := range m.Movers {
+		if s := m.String(); s != EmptyObjectStr {
+			ms = append(ms, m.String())
+		}
+	}
+	if len(ms) < 1 {
```
antas-marcin, line 37: There are quite a lot of if statements that are testing certain scenarios, which I think can be ommited, you can just build the parts of the argument (concepts, force, objects) and then just at the very end concatenate the results into one string as we do this in other argument builders

v4/weaviate/graphql/neartextbuilder.go
```diff
@@ -1,11 +1,72 @@
...
+// MoverObject is the object the search is supposed to move close to (or further away from) it.
+type MoverObject struct {
+	ID     string
+	Beacon string
+}
+
+//String returns string representation of m as {"id": "value" beacon:"value"}.
+// Empty fields are considered optional and are excluded.
+// It returns EmptyObjectStr if both fields are empty
+func (m *MoverObject) String() string {
```
antas-marcin, line 57: I think that this method can be made private

v4/weaviate/graphql/neartextbuilder.go
redouan-rhazouani: I think on the contrary `marshalStrings` makes code simpler and better to read and refactor:

I can think at leat of two reasons why this is the case:
1. `marshalStrings` has the single responsibility of marshalling slices of strings which results in better readability and testability.
2. This function is need in different places which removes redundancy. In contrast the proposed solution would result in redundant and hard to test code.

This function replaces ` concepts, _ := json.Marshal(m.Concepts)` which as @parkerduckworth has noticed ignore the error.

v4/weaviate/graphql/neartextbuilder.go
redouan-rhazouani: Do you mean changing MoverObject to MoverObjects?

v4/weaviate/graphql/neartextbuilder.go
redouan-rhazouani: The "}" is included in the next line

v4/weaviate/graphql/neartextbuilder.go
redouan-rhazouani: I am not quite sure what can be omitted. 

However the purpose of the statement above is to exclude the moverObjects when there is no mover specified

v4/weaviate/graphql/neartextbuilder.go
redouan-rhazouani: Done

v4/weaviate/graphql/neartextbuilder.go
```diff
@@ -1,11 +1,72 @@
...
+	// Objects used to adjust the serach direction
+	Objects []MoverObject
+}
+
+func (m *MoveParameters) String() string {
+	concepts := marshalStrings(m.Concepts)
+	ms := make([]string, 0, len(m.Objects))
+	for _, m := range m.Objects {
+		if s := m.String(); s != EmptyObjectStr {
+			ms = append(ms, m.String())
```
antas-marcin, line 34: `m.String()` -> `s`

v4/weaviate/graphql/neartextbuilder.go
```diff
@@ -1,11 +1,72 @@
...
+	return fmt.Sprintf("%s %s: %v %s: %v}", s, fldMoverForce, m.Force, fldMoverObjects, ms)
+}
+
+// MoverObject is the object the search is supposed to move close to (or further away from) it.
+type MoverObject struct {
+	ID     string
+	Beacon string
+}
+
+//String returns string representation of m as {"id": "value" beacon:"value"}.
```
antas-marcin, line 54: `//String` -> `// String`

v4/weaviate/graphql/neartextbuilder.go
redouan-rhazouani: Nice catch done

v4/weaviate/graphql/neartextbuilder.go
redouan-rhazouani: makes sense done

PR #79, opened 2022-06-14
Feature/weaviate 125

parkerduckworth:
This PR contains updates for the following endpoints:

- `GET /v1/objects/{className}/{id}`
- `DELETE /v1/objects/{className}/{id}`
- `PUT /v1/objects/{className}/{id}`
- `PATCH /v1/objects/{className}/{id}`
- `HEAD /v1/objects/{className}/{id}`

### DIFF
--- a/v4/weaviate/data/objectsGetter.go
+++ b/v4/weaviate/data/objectsGetter.go
@@ -15,15 +15,22 @@ import (
 type ObjectsGetter struct {
 	connection           *connection.Connection
 	id                   string
+	className            string
 	additionalProperties []string
 	withLimit            bool
 	limit                int
 }
 
-// WithID specifies the uuid of the Thing that should be retrieved
-//  if omitted a set of objects matching the builder specifications will be retrieved
+// WithID specifies the uuid of the object that should be retrieved
+// if omitted a set of objects matching the builder specifications will be retrieved
 func (getter *ObjectsGetter) WithID(id string) *ObjectsGetter {
 	getter.id = id
+	return getter
+}
+
+// WithClassName specifies the class name of the object that should be retrieved
+func (getter *ObjectsGetter) WithClassName(className string) *ObjectsGetter {
+	getter.className = className
 	return getter
 }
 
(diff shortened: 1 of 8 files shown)


v4/weaviate/data/path.go
```diff
@@ -0,0 +1,11 @@
+package data
+
+import "path"
```
antas-marcin, line 3: I didn't know this package, nice!

redouan-rhazouani (approved):

v4/weaviate/data/objectsGetter.go
```diff
@@ -48,29 +55,47 @@ func (getter *ObjectsGetter) WithLimit(limit int) *ObjectsGetter {
...
+	}
+	return responseData, err
+}
+
+func (getter *ObjectsGetter) buildPath() string {
+	basePath := buildObjectsPath(getter.id, getter.className)
+
+	params := buildAdditionalParams(getter.additionalProperties)
+	if getter.withLimit {
+		params += fmt.Sprintf("?limit=%v", getter.limit)
```
aliszka, line 91: if additional params exists, would the final params string be `?include=xxx?limit=yyy` instead `?include=xxx&limit=yyy`?

v4/weaviate/data/objectsGetter.go
```diff
@@ -48,29 +55,47 @@ func (getter *ObjectsGetter) WithLimit(limit int) *ObjectsGetter {
...
 }
 
-// getAdditionalParams build the query URL parameters for the requested underscore properties
-func getAdditionalParams(additionalProperties []string) string {
+func (getter *ObjectsGetter) objectList(ctx context.Context) (*connection.ResponseData, error) {
+	responseData, err := getter.connection.RunREST(ctx, getter.buildPath(), http.MethodGet, nil)
+	if err != nil {
+		return responseData, except.NewDerivedWeaviateClientError(err)
+	}
+	return responseData, err
```
aliszka, line 83: maybe `return responseData, nil` for easier reading?

v4/weaviate/data/objectsGetter.go
antas-marcin: good catch! looks like this implementation never worked properly if additional param was passed with limit

PR #87, opened 2022-07-18
Classification tests with where filters set

antas-marcin:


v4/test/classifications/classification_with_where_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+	t.Run("POST /classifications", func(t *testing.T) {
+		client := testsuit.CreateTestClient()
+		createClassificationClasses(t, client)
+
+		sourceWhere := filters.Where().
+			WithOperator(filters.Like).
+			WithPath([]string{"id"}).
+			WithValueString("*")
+		classifyProperties := []string{"taged"}
```
parkerduckworth, line 34: `taged` -> maybe meant to be `tagged`?

parkerduckworth (approved):

v4/test/classifications/classification_with_where_test.go
antas-marcin: Actually that's the name of property tak we use in our tests. I have just re-used that.
Let me check if it would be a quick to fix that property name.

antas-marcin: OK. I have fixed all of the occurrences of this property

PR #94, opened 2022-09-12
Add the ability to pass raw string graph Ql queries in go clients

donomii:


v4/weaviate/graphql/rawgql_test.go
```diff
@@ -0,0 +1,30 @@
...
+			},
+			want: "query { Get { Things { Thing { name } } } }",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			e := &RawGraphQLQueryBuilder{query: tt.fields.query}
+			if got := e.build(); got != tt.want {
+				t.Errorf("RawGQLBuilder.build() = %v, want %v", got, tt.want)
+			}
```
parkerduckworth, line 27: Would be good to have some error cases added

parkerduckworth (changes requested):
I think we should have some integration tests using raw string queries as well

v4/weaviate/graphql/rawgql_test.go
donomii: Integration tests added

donomii (commented):
Something odd here, I have to review my own changes?

v4/weaviate/graphql/gql.go
```diff
@@ -34,6 +34,11 @@ func (api *API) Aggregate() *AggregateBuilder {
 	return &AggregateBuilder{connection: api.connection}
 }
 
+// RawGraphQLQueryBuilder creates a raw GraphQL query
+func (api *API) Raw(query string) *RawGraphQLQueryBuilder {
+	return &RawGraphQLQueryBuilder{connection: api.connection, query: query}
```
aliszka, line 39: could you please rename struct to `Raw` (eventually `RawBuilder`, though there is not much to be build) to match other clients (`GetBuilder`, `Explore`, `AggregateBuilder`)? and filename to match new name.

aliszka: query should be added by `WithQuery` method, not given in the `constructor` to be consistent with other clients

v4/weaviate/graphql/rawgql.go
```diff
@@ -0,0 +1,23 @@
...
+	query      string
+}
+
+// Do execute the GraphQL query
+func (gql *RawGraphQLQueryBuilder) Do(ctx context.Context) (*models.GraphQLResponse, error) {
+	return runGraphQLQuery(ctx, gql.connection, gql.query)
+}
+
+// return the query string
+func (gql *RawGraphQLQueryBuilder) build() string {
```
aliszka, line 21: is it used somewhere?

v4/test/graphql/graphql_test.go
```diff
@@ -16,6 +17,37 @@ import (
...
+		})
+
+		t.Run("get all", func(t *testing.T) {
+			resultSet, gqlErr := client.GraphQL().Raw("{Get {Pizza {name}}}").Do(context.Background())
+			assert.Nil(t, gqlErr)
+
+			get := resultSet.Data["Get"].(map[string]interface{})
+			pizza := get["Pizza"].([]interface{})
+			assert.Equal(t, 4, len(pizza))
+		})
```
antas-marcin, line 41: You can actually just move this test inside `t.Run("Get")` there's already a DB being created and populated no need to do this once again in a separate test

v4/weaviate/graphql/rawgql.go
donomii: yes

v4/test/graphql/graphql_test.go
antas-marcin: @donomii you want to leave this one small separately in `t.Run("RawGraphQL",`, yes?

antas-marcin: bc I suggested that it's so small that you can just move it inside the `t.Run("Get")` suite to be run there

v4/weaviate/graphql/gql.go
antas-marcin: Good point to make it concise with other clients, so that the `Raw` query invocation would look similar to other clients, like this:
```
client.GraphQL().Raw().WithQuery(...).Do()
```

v4/weaviate/graphql/rawgql.go
```diff
@@ -0,0 +1,29 @@
...
+
+// Raw for accepting a prebuilt query from the user
+type Raw struct {
+	connection rest
+	query      string
+}
+
+// Do execute the GraphQL query
+func (gql *Raw) Do(ctx context.Context) (*models.GraphQLResponse, error) {
+	return runGraphQLQuery(ctx, gql.connection, gql.query)
```
antas-marcin, line 17: to be concise I would use `build()` method in `runGraphQLQuery` instead of `qql.query`

v4/weaviate/graphql/rawgql_test.go
```diff
@@ -0,0 +1,30 @@
+package graphql
+
+import "testing"
+
+func TestRawGQLBuilder_build(t *testing.T) {
```
antas-marcin, line 5: In addition to this test there was also a journey test which now disappeared... can you add back this journey test also?

PR #101, opened 2022-12-01
Add explicit support for User/PW, client credentials and Bearer token oidc auth flows

dirkkul:
… oidc auth flows

Before you could use all (most?) oauth methods by creating a `oauth2.NewClient` [as described in the docs](https://weaviate.io/developers/weaviate/current/client-libraries/go.html#authentication). However, creating this client isn't straight forward and you need to supply configuration values that we can get automatically from the weaviate oidc configuration.
(Before looking into the code I wasn't aware that this includes authentication methods other than supplying a token). 

This PR adds a new `NewConfig` function, that creates an oauth2 client as part of the client configuration:

`NewConfig(Host string, Scheme string, AuthConfig auth.Config, Headers map[string]string)`

The possibilities for auth.Config are:
- `auth.BearerToken{AccessToken:*token*, ExpiresIn: *seconds*, RefreshToken:*token*}`
- `auth.ResourceOwnerPasswordFlow{Username: *user*, Password: *pw*, scope:*scopes*}`
- `auth.ClientCredentials{ClientSecret: *secret*, scopes: *scopes*}`

To get a client with authentication enabled, use:
```
clientCredentialConf := auth.ResourceOwnerPasswordFlow{Username: "*user*", Password: *pw*}
cfg, err := weaviate.NewConfig("localhost:"+fmt.Sprint(WCSPort), "http", clientCredentialConf, nil)
client := weaviate.New(*cfg)
```

and without:
```
cfg, err := weaviate.NewConfig("localhost:"+fmt.Sprint(WCSPort), "http", nil, nil)
client := weaviate.New(*cfg)
```

There are tests that test authenticating with a local weaviate and WCS, Okta and Azure as authentication providers.


v4/weaviate/auth/auth.go
```diff
@@ -0,0 +1,119 @@
...
+import (
+	"context"
+	"errors"
+	"fmt"
+	"github.com/semi-technologies/weaviate-go-client/v4/weaviate/connection"
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/clientcredentials"
+	"log"
+	"net/http"
+	"strings"
```
parkerduckworth, line 12: sort these please :)

v4/weaviate/auth/auth.go
```diff
@@ -0,0 +1,119 @@
...
+
+type Config interface {
+	GetAuthClient(con *connection.Connection) (*http.Client, error)
+}
+
+type authBase struct {
+}
+
+func (ab authBase) getIdAndTokenEndpoint(con *connection.Connection) (string, string, error) {
+	rest, err := con.RunREST(context.TODO(), "/.well-known/openid-configuration", http.MethodGet, nil)
```
parkerduckworth, line 25: Can just use the `OidcConfigURL` const above

v4/weaviate/auth/auth.go
```diff
@@ -0,0 +1,119 @@
...
+	if err != nil {
+		return "", "", err
+	}
+	if rest.StatusCode == 404 {
+		log.Println("The client was configured to use authentication, but weaviate is configured without authentication. Are you sure this is correct?")
+		return "", "", nil
+	} else if rest.StatusCode != 200 {
+		return "", "", errors.New("OIDC configuration url " + OidcConfigURL + "returned status code " + fmt.Sprint(rest.StatusCode))
+	}
```
parkerduckworth, line 34: would be more readable as a `switch`/`case` IMO

v4/weaviate/auth/auth.go
```diff
@@ -0,0 +1,119 @@
...
+	var result map[string]interface{}
+	decodeErr := rest.DecodeBodyIntoTarget(&result)
+	if decodeErr != nil {
+		return "", "", decodeErr
+	}
+
+	clientId := result["clientId"].(string)
+	href := result["href"].(string)
+	endpoints, err := con.RunRESTExternal(context.TODO(), href, http.MethodGet, nil)
```
parkerduckworth, line 43: can unmarshal into a struct instead:

```go
cfg := struct {
    Href       string `json:"href"`
    ClientID string `json:"clientId"`
}{}
decodeErr := rest.DecodeBodyIntoTarget(&cfg)
if decodeErr != nil {
	return "", "", decodeErr
}

clientId := cfg.ClientID
href := href := cfg.Href
```

v4/test/auth/auth_test.go
```diff
@@ -0,0 +1,155 @@
...
+	"bytes"
+	"context"
+	"fmt"
+	"github.com/semi-technologies/weaviate-go-client/v4/weaviate"
+	"github.com/semi-technologies/weaviate-go-client/v4/weaviate/auth"
+	"github.com/stretchr/testify/assert"
+	"log"
+	"os"
+	"strings"
+	"testing"
```
parkerduckworth, line 13: sort please :)

v4/weaviate/auth/auth.go
```diff
@@ -0,0 +1,119 @@
...
+func (ab authBase) getIdAndTokenEndpoint(con *connection.Connection) (string, string, error) {
+	rest, err := con.RunREST(context.TODO(), "/.well-known/openid-configuration", http.MethodGet, nil)
+	if err != nil {
+		return "", "", err
+	}
+	if rest.StatusCode == 404 {
+		log.Println("The client was configured to use authentication, but weaviate is configured without authentication. Are you sure this is correct?")
+		return "", "", nil
+	} else if rest.StatusCode != 200 {
+		return "", "", errors.New("OIDC configuration url " + OidcConfigURL + "returned status code " + fmt.Sprint(rest.StatusCode))
```
antas-marcin, line 33: would be great to use `fmt.Errorf(...)` instead of `errors.New`

v4/weaviate/auth/auth.go
```diff
@@ -0,0 +1,119 @@
...
+	ClientSecret string
+	Scopes       []string
+	authBase
+}
+
+func (cc ClientCredentials) GetAuthClient(con *connection.Connection) (*http.Client, error) {
+	clientId, tokenEndpoint, err := cc.getIdAndTokenEndpoint(con)
+	if err != nil {
+		return nil, err
+	} else if err == nil && clientId == "" && tokenEndpoint == "" {
```
antas-marcin, line 65: you can get rid of `err == nil` condition here

v4/weaviate/auth/auth.go
```diff
@@ -0,0 +1,119 @@
...
+	Username string
+	Password string
+	authBase
+}
+
+func (ro ResourceOwnerPasswordFlow) GetAuthClient(con *connection.Connection) (*http.Client, error) {
+	clientId, tokenEndpoint, err := ro.getIdAndTokenEndpoint(con)
+	if err != nil {
+		return nil, err
+	} else if err == nil && clientId == "" && tokenEndpoint == "" {
```
antas-marcin, line 94: you can omit `err == nil` condition here

v4/weaviate/auth/auth.go
```diff
@@ -0,0 +1,119 @@
...
+
+type BearerToken struct {
+	Token string
+	authBase
+}
+
+func (bt BearerToken) GetAuthClient(con *connection.Connection) (*http.Client, error) {
+	// we don't need these values, but we can check if weaviate is configured with authentication enabled
+	clientId, tokenEndpoint, err := bt.getIdAndTokenEndpoint(con)
+	if err == nil && clientId == "" && tokenEndpoint == "" {
```
antas-marcin, line 114: why you're not checking here if `err != nil'?

v4/weaviate/auth/auth.go
```diff
@@ -0,0 +1,119 @@
...
+	"fmt"
+	"github.com/semi-technologies/weaviate-go-client/v4/weaviate/connection"
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/clientcredentials"
+	"log"
+	"net/http"
+	"strings"
+)
+
+const OidcConfigURL = "/.well-known/openid-configuration"
```
antas-marcin, line 15: is this const used outside the package? if not then it can be defined as a private variable, with a lower letter

v4/weaviate/auth/auth.go
dirkkul: See comment right above :)

v4/weaviate/weaviate_client.go
```diff
@@ -34,6 +35,19 @@ type Config struct {
 	Headers map[string]string
 }
 
+func NewConfig(Host string, Scheme string, AuthConfig auth.Config, Headers map[string]string) (*Config, error) {
+	var client *http.Client = nil
+	var err error
+	if AuthConfig != nil {
```
antas-marcin, line 41: I would make `AutoConfig` and `Headers` variable names to start with lower letter

v4/weaviate/weaviate_client.go
```diff
@@ -34,6 +35,19 @@ type Config struct {
 	Headers map[string]string
 }
 
+func NewConfig(Host string, Scheme string, AuthConfig auth.Config, Headers map[string]string) (*Config, error) {
```
antas-marcin, line 38: I would make all variable names to start with lower letter

v4/weaviate/testenv/environment.go
```diff
@@ -42,7 +44,7 @@ func SetupLocalWeaviateDeprecated() error {
 	return WaitForWeaviate()
 }
 
-func isExternalWeaviateRunning() bool {
+func IsExternalWeaviateRunning() bool {
```
antas-marcin, line 47: is this function being used somewhere outside this package? if not then I would make it private

v4/weaviate/auth/auth.go
```diff
@@ -21,21 +21,22 @@ type Config interface {
...
 	if err != nil {
-		return "", "", err
+		return "", []string{}, "", err
 	}
 	cfg := struct {
-		Href     string `json:"href"`
-		ClientID string `json:"clientId"`
+		Href     string   `json:"href"`
+		ClientID string   `json:"clientId"`
+		Scopes   []string `json:"scopes"`
```
parkerduckworth, line 32: It would probably be better to define this struct outside the scope of this function, and use it as the return type. It is not very clear what each return type in the signature is supposed to be

Either that, or name the return params

v4/weaviate/auth/auth.go
```diff
@@ -46,22 +47,22 @@ func (ab authBase) getIdAndTokenEndpoint(ctx context.Context, con *connection.Co
...
+		return "", []string{}, "", err
 	}
 	var resultEndpoints map[string]interface{}
 	decodeEndpointErr := endpoints.DecodeBodyIntoTarget(&resultEndpoints)
 	if decodeEndpointErr != nil {
-		return "", "", err
+		return "", []string{}, "", err
 	}
-	return cfg.ClientID, resultEndpoints["token_endpoint"].(string), nil
+	return cfg.ClientID, cfg.Scopes, resultEndpoints["token_endpoint"].(string), nil
```
parkerduckworth, line 65: Maybe want to do a type assertion check on `resultEndpoints["token_endpoint"]`, otherwise if this panics it may cause a bad UX

dirkkul: done

v4/weaviate/auth/auth.go
dirkkul: went a different route, let me know what you think

parkerduckworth: great, this is also a good solution.

v4/test/auth/auth_test.go
```diff
@@ -0,0 +1,257 @@
...
+	"io"
+	"log"
+	"net/http"
+	"net/url"
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/weaviate/weaviate-go-client/v4/test/testsuit"
+
```
parkerduckworth, line 17: rm imports newline. maybe we can add a linter to the ci pipeline

v4/test/testsuit/generics.go
```diff
@@ -2,12 +2,26 @@ package testsuit
 
 import (
 	"context"
+	"fmt"
+	"net/http"
+	"os"
 	"testing"
 
+	"github.com/weaviate/weaviate-go-client/v4/weaviate/auth"
+
```
parkerduckworth, line 11: rm imports newline

v4/weaviate/connection/rest.go
```diff
@@ -4,11 +4,11 @@ import (
...
 	"context"
 	"encoding/json"
+	"github.com/weaviate/weaviate-go-client/v4/weaviate/fault"
+	"golang.org/x/oauth2"
 	"io"
-	"io/ioutil"
 	"net/http"
-
-	"github.com/weaviate/weaviate-go-client/v4/weaviate/fault"
+	"time"
```
parkerduckworth, line 11: sort imports

v4/weaviate/connection/rest.go
```diff
@@ -27,12 +27,40 @@ func NewConnection(scheme string, host string, httpClient *http.Client, headers
...
+			if token.RefreshToken == "" {
+				return
+			}
+
+			timeToSleep := token.Expiry.Sub(time.Now()) - time.Second/10
+			time.Sleep(timeToSleep)
+			_, err = connection.RunREST(context.TODO(), "/meta", http.MethodGet, nil)
+			if err != nil {
+				return
+			}
```
parkerduckworth, line 59: Maybe we would want to log or propagate these errors

v4/weaviate/weaviate_client.go
```diff
@@ -2,6 +2,7 @@ package weaviate
 
 import (
 	"context"
+	"github.com/weaviate/weaviate-go-client/v4/weaviate/auth"
```
parkerduckworth, line 5: should be a newline separating stdlib imports from the rest

v4/weaviate/weaviate_client.go
```diff
@@ -33,6 +34,19 @@ type Config struct {
 	Headers map[string]string
 }
 
+func NewConfig(host string, scheme string, authConfig auth.Config, headers map[string]string) (*Config, error) {
+	var client *http.Client = nil
```
parkerduckworth, line 38: if you declare a pointer type, it will already hold a value of `nil` by default, no need to manually assign:
```
var client *http.Client
```

v4/weaviate/connection/rest.go
dirkkul: added logging

v4/test/auth/auth_mock_test.go
```diff
@@ -0,0 +1,267 @@
...
+	defer s.Close()
+
+	cfg, err := weaviate.NewConfig(strings.TrimPrefix(s.URL, "http://"), "http", auth.ClientCredentials{ClientSecret: "SecretValue"}, nil)
+	assert.Nil(t, err)
+	client := weaviate.New(*cfg)
+	AuthErr := client.Schema().AllDeleter().Do(context.TODO())
+	assert.Nil(t, AuthErr)
+	assert.Equal(t, i, 3) // client does 3 initial calls to token endpoint
+
+	time.Sleep(time.Second * 5)
```
antas-marcin, line 124: I would suggest to change this gouroutine and refactor it using a `ticker` not `sleeps`

v4/weaviate/weaviate_client.go
dirkkul: changed

v4/weaviate/connection/rest.go
```diff
@@ -27,12 +29,43 @@ func NewConnection(scheme string, host string, httpClient *http.Client, headers
...
+			// there is no point in manual refreshing if there is no refresh token
+			if token.RefreshToken == "" {
+				log.Println("No refresh token, exiting background goroutine")
+				return
+			}
+
+			timeToSleep := token.Expiry.Sub(time.Now()) - time.Second/10
+			time.Sleep(timeToSleep)
```
antas-marcin, line 59: I would suggest to refactor this goroutine to use channels with ticker instead of sleep. it would make also a code a little bit more readable

dirkkul: refactored the goroutine a bit, please have another look

v4/test/auth/auth_test.go
parkerduckworth: gofumpt will only properly sort if there are no line breaks in the import block 🙃 but good to know you are tracking a linter issue.

v4/weaviate/connection/rest.go
```diff
@@ -77,7 +132,7 @@ func (con *Connection) createRequest(ctx context.Context, path string,
 		return nil, err
 	}
 	con.addHeaderToRequest(request)
-	//con.addURLParametersToRequest(request, urlParameters)
+	// con.addURLParametersToRequest(request, urlParameters)
```
parkerduckworth, line 135: can this be removed?

v4/test/auth/auth_test.go
dirkkul: ok, now

v4/weaviate/connection/rest.go
parkerduckworth: Hmm I'm not seeing any logging.  What I mean is that the goroutine should be logging on errors. So `if err != nil`, log the error message before returning from the goroutine

v4/weaviate/connection/rest.go
dirkkul: That was there before, but I'll delete it. Nobody seems to miss it and should be easy to readd

parkerduckworth: Ah , ok 👍

v4/weaviate/connection/rest.go
dirkkul: There definitely is logging, are you loocking at the [new code](https://github.com/weaviate/weaviate-go-client/blob/b35f005a054992a48f950ab554b36d01bdc24be4/v4/weaviate/connection/rest.go#L84)? The snippet here is not updated

parkerduckworth: Ah weird, I was looking at the refreshed changes [here](https://github.com/weaviate/weaviate-go-client/pull/101/files), but I guess they were still outdated. Ok LGTM

parkerduckworth (approved):

antas-marcin (approved):

PR #102, opened 2022-12-13
Add support for node_name and consistency_level get by id query params

parkerduckworth:


v4/weaviate/data/objectsGetter.go
```diff
@@ -113,6 +135,13 @@ func (getter *ObjectsGetter) buildPathParams() string {
 		}
 	}
 
+	if getter.withConsistencyLevel {
+		pathParams = append(pathParams, fmt.Sprintf("consistency_level=%v", getter.consistencyLevel))
+	}
+	if getter.withNodeName {
+		pathParams = append(pathParams, fmt.Sprintf("node_name=%v", getter.nodeName))
+	}
```
dirkkul, line 143: Is there any validation that only one of both is added?

v4/weaviate/data/objectsGetter.go
```diff
@@ -20,6 +20,10 @@ type ObjectsGetter struct {
 	additionalProperties []string
 	withLimit            bool
 	limit                int
+	withConsistencyLevel bool
+	consistencyLevel     string
+	withNodeName         bool
+	nodeName             string
```
dirkkul, line 26: instead of using two additional bools, wouldn't `consistencyLevel/nodeName != "" ` equivalent?

v4/weaviate/data/objectsGetter.go
```diff
@@ -55,6 +59,24 @@ func (getter *ObjectsGetter) WithLimit(limit int) *ObjectsGetter {
 	return getter
 }
 
+// WithConsistencyLevel determines how many replicas must acknowledge a request
+// before it is considered successful. Mutually exclusive with node_name param.
+// Can be one of 'ALL', 'ONE', or 'QUORUM'. Note that WithConsistencyLevel and
+// WithNodeName are mutually exclusive.
+func (getter *ObjectsGetter) WithConsistencyLevel(cl string) *ObjectsGetter {
```
dirkkul, line 66: Should we add validation for the cl that is passed in? Or does Weaviate return an error message?

dirkkul (approved):

v4/weaviate/data/objectsGetter.go
parkerduckworth: If you prefer I can do so, but just note that the rest of the client uses this pattern

v4/weaviate/data/objectsGetter.go
parkerduckworth: Correct, this is validated sever-side

v4/weaviate/data/objectsGetter.go
parkerduckworth: Yes, the server validates this

v4/weaviate/data/objectsGetter.go
antas-marcin: hmm, actually in this case we could rely on `consistencyLevel/nodeName != ""` check, those boolean flags were introduced to determine if the value was set, for example `limit int` field, with that type of var we need to have a `withLimit bool` to determine if the user for example explicitly have set `0` value

antas-marcin: but here with strings I think we can skip it

dirkkul: removed them

PR #103, opened 2022-12-14
Add bm25 and hybrid support to Go client

donomii:


v4/weaviate/graphql/hybridbuilder.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+// WithQuery the search string
+func (e *HybridArgumentBuilder) WithQuery(query string) *HybridArgumentBuilder {
+	e.Query = query
+	return e
+}
+
+// WithVector the vector.  Can be omitted
+func (e *HybridArgumentBuilder) WithVector(vector ...float32) *HybridArgumentBuilder {
```
antas-marcin, line 21: shouldn't a vector argument be declared as `vector []float32` but not in dot notation? (that would be consistent with other arguments)

donomii: That sounds good

Issue #49, opened 2022-03-24
Support querying with multiple classes at once


parkerduckworth:
Currently querying with only one class is supported. Queries like this are not currently possible:

```
{
  Get {
    ClassOne {
      field
    }
    ClassTwo {
      field
    }
  }
}
```

---

The change should be relatively simple.

The `GetBuilder` type can only support one class by design:

```
type GetBuilder struct {
	connection rest
	className  string
	withFields string

	includesFilterClause bool // true if brackets behind class is needed
	withWhereFilter      string
	includesLimit        bool
	limit                int
	includesOffset       bool
	offset               int
	withNearTextFilter   *NearTextArgumentBuilder
	withNearVectorFilter string
	withNearObjectFilter *NearObjectArgumentBuilder
	withGroupFilter      string
	withAskFilter        *AskArgumentBuilder
	withNearImageFilter  *NearImageArgumentBuilder
}
```

So each time you call `WithClassName`, (or any other query modifier), it’s just resetting the `GetBuilder.className` field. What needs to change is that the GetBuilder can look something like this:

```
type GetBuilder struct {
	ClassBuilders []*ClassBuilder
}

type ClassBuilder struct {
	connection rest
	className  string
	withFields string

	includesFilterClause bool // true if brackets behind class is needed
	withWhereFilter      string
	includesLimit        bool
	limit                int
	includesOffset       bool
	offset               int
	withNearTextFilter   *NearTextArgumentBuilder
	withNearVectorFilter string
	withNearObjectFilter *NearObjectArgumentBuilder
	withGroupFilter      string
	withAskFilter        *AskArgumentBuilder
	withNearImageFilter  *NearImageArgumentBuilder
}
```

shreyanshshah27:
Hey @parkerduckworth, I'm new to this repo/project and wanted to get idea about this code-base and eventually start contributing. This seems like a good entry point for me. So I just wanted to confirm whether this issue is still open or this has been already resolved.

parkerduckworth:
Hey @shreyanshshah27 this issue is still open and available 🙂

shreyanshshah27:
Thanks for your quick response @parkerduckworth. I've tried to add this feature but instead of using `[]*ClassBuilder` as suggested by you, I am using `map[string]*ClassBuilder`, and in the long run I believe this will be a better solution. Requesting you and all other contributors to review this PR and provide their comments and inputs.

dirkkul:
[...]

PR #106, opened 2023-01-08
gh-49 support for querying multiple classes

shreyanshshah27:


shreyanshshah27:
I agree with the CLA.

v4/weaviate/graphql/get_test.go
```diff
@@ -79,10 +82,16 @@ func TestQueryBuilder(t *testing.T) {
...
+		query := builder.AddQueryClass(
+			NewQueryClassBuilder("Pizza").
+				WithFields(name).
+				WithWhere(where),
+		).build()
+
+		// query := builder.WithClassName("Pizza").
+		// 	WithFields(name).
+		// 	WithWhere(where).
+		// 	build()
```
dirkkul, line 94: remove this?

v4/weaviate/graphql/get.go
```diff
@@ -32,168 +36,175 @@ type GetBuilder struct {
 	withHybrid           *HybridArgumentBuilder
 }
 
-// WithClassName that should be queried
-func (gb *GetBuilder) WithClassName(name string) *GetBuilder {
-	gb.className = name
-	return gb
+// ClassName that should be queried
+func NewQueryClassBuilder(className string) *ClassBuilder {
```
dirkkul, line 40: This is a breaking change, right? And all users would need to update their implementation?

Wouldn't be a big fan of this. Can we keep the old behavior and add the possiblity to have multiple classes in addition? @parkerduckworth what do you think?

v4/weaviate/graphql/get.go
```diff
@@ -32,168 +36,175 @@ type GetBuilder struct {
...
-	if gb.includesFilterClause {
-		filterClause = gb.createFilterClause()
+	var query string
+	for _, class := range gb.ClassBuilders {
+		filterClause := ""
+		if class.includesFilterClause {
+			filterClause = gb.createFilterClause(class)
+		}
+		fieldsClause := gb.createFieldsClause(class)
+		query = query + fmt.Sprintf("%v %v {%v}", class.className, filterClause, fieldsClause) + " "
```
dirkkul, line 155: use `query +=`

dirkkul (changes requested):

v4/weaviate/graphql/get_test.go
shreyanshshah27: Fixed.

v4/weaviate/graphql/get.go
shreyanshshah27: Fixed.

v4/weaviate/graphql/get.go
dirkkul: that looks much better :)

I noticed that it duplicates a lot of the code for GetBuilder, I think we can improve that.

How about doing:
- in get.go:
```
type GetBuilder struct {
	connection rest
        GetBuilderBase
}

type GetBuilderBase struct{
	className  string
        ..other members...
}
```
- in multiclassbuilder.go
```
type MultiClassBuilder struct {
	connection    rest
	classBuilders map[string]*GetBuilderMulti
}

type GetBuilderMulti struct {
        GetBuilderBase
}

...additional member functions....
```
(not tested, just typed here)

A few more tests would be great, though

shreyanshshah27: I've tried to incorporate your suggestions but in a slightly different way. Let me know your thoughts. If this looks good, will add tests for `MultiClassBuilder `.

dirkkul: Hi,

now we have the problem again, that we have a breaking change :/ Sorry, didn't think of that.

New proposal:
How about we leave the `GetBuilder` class like it is. And then have the following?
```
type MultiClassBuilder struct {
	connection    rest
	classBuilders map[string]*GetBuilder
}

NewMultiClassBuilder(connection rest){
return MultiClassBuilder{....}
}

func (m *MultiClassBuilder) AddClass(className string) *GetBuilder{
  m.classBuilders[className] = GetBuilder{conn: m.connection, className: className}
  return m.classBuilders[className]
}

func (m *MultiClassBuilder) ModifyClass(className string) *GetBuilder{
  return m.classBuilders[className]
}
```

Then we're backwards compatible and have a nice interface to add multiple classes. What do you think? Hope I didn't miss anything :)

Instead of returning the GetBuilder, we could also go for something that can be chained with multiple classes

shreyanshshah27: Hey @dirkkul ,

Sorry I couldn't understand how this is a breaking change. The only thing we are changing is the underlying structure of `client.GraphQL().Get()`, the fields defined on this weren't even earlier exposed/accessible and all the methods are still defined as they were earlier.

It would be great if you can add an example demonstrating code breaking.

dirkkul: Hi,

before users would write

```
		builder := GetBuilder{
			connection:           conMock,
			..other options...
		}
```

now they need to write the following and update all existing code.
```
GetBuilder{
			connection: conMock,
			builderBase: builderBase{
				includesFilterClause: false,
			},
		}
```

Or did I miss something? At least you updated a few tests that use the GetBuilder.

shreyanshshah27: I think you misunderstood, the file that you're referring for changes includes unit tests. Yes that is changed. But end user actually won't be using it this way, right? 

The way users will be using isn't changes. Even before users were doing this and even after this change they will be doing :

`
client.GraphQL().Get().WithClassName("Pizza").WithFields(name).Do(context.Background())
`

You can refer to weaviate-go-client/v4/test/grapgql/graphql_test.go file for more example.

shreyanshshah27: @dirkkul, I thing this would help you understand my thoughts,

You are only half-correct when you say users can write this : 
```
builder := GetBuilder{
	connection:           conMock,  
	..other options...  
}
```
Outside `weaviate-go-client/v4/weaviate`, users will only be able to write this :
```
builder := graphql.GetBuilder{}
```
Outside the weaviate pkg, users won't be able to set `connection` or any other field of `GetBuilder` structure.

Let's extent this argument, assuming user is writing something like this 
```
bd := graphql.GetBuilder{}
bd.WithClassName("Pizza").WithAsk(...)
```
and since there is no way to pass variable `bd` as argument to some function and execute the get query, the only way user can execute this query is by doing this : 
```
bd.Do(context.Background())
```
and this is gonna return some error, since `connection` field isn't set. So logically speaking there is no point of getting a `graphql.GetBuilder` variable without using `client.GraphQL().Get()` method.

So yes, you are correct that user will have to re-write/update existing code if they are declaring `graphql.GetBuilder` variable without using `client.GraphQL().Get()` method, but I believe this is very less likely that someone will be doing it the wrong way, since there is no point of doing it.

Please correct me, if I am wrong.

v4/weaviate/graphql/multiclassbuilder.go
```diff
@@ -0,0 +1,190 @@
...
+		for i := range class.withFields {
+			fields[i] = class.withFields[i].build()
+		}
+		return strings.Join(fields, " ")
+	}
+	return ""
+}
+
+// WithFields included in the result set
+func (bb *builderBase) WithFields(fields ...Field) *builderBase {
```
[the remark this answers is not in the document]
shreyanshshah27: Yes you are correct we should avoid defining duplicate methods. But if we remove it from `GetBuilder` and keep it only on `builderBase` then it would be a breaking change. So on the `GetBuilder` structure we can only change underlying structure not the methods defined on it. 

We need to define these methods on `builderBase` because in `MultiClassBuilder` structure, underlying we are using `builderBase`.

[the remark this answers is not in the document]
dirkkul: I had a bit more time today and played a bit around with your code.
Thanks to your explanation above I made a few changes that reduce the amount of duplicate code, have a look at this branch: https://github.com/semi-technologies/weaviate-go-client/tree/pr/106.

What do you think?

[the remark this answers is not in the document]
shreyanshshah27: I had similar approach in mind but I wanted to keep `MultiClassBuilder` loosely coupled with `GetBuilder`. Doing this is exactly opposite of that, and few other reasons are as:

1. Earlier there was only one way to get `GetBuilder` but now since `MultiClassBuilder` also refers to `GetBuilder`, user can do something like this : 
```
builder := graphql.NewQueryClassBuilder("Pizza")
```
and it will return the variable of type `*graphql.GetBuilder` and on this `Do()` method is also defined. Even though calling it will only return error, but it makes it a bit confusing.

2. Now in `MultiClassBuilder` struct we have multiple connection available. One in `MultiClassBuilder` itself and every time we add a new class to `MultiClassBuilder`, that class will have one connection available. This won't harm in anyway, but it doesn't make sense and it's just not good design.

We can go ahead with this approach but instead of solving legacy design problems, we'll just be adding more bad design and in future it while trying to solve this, it'll become breaking changes. So I would prefer having some duplication over adding confusing design.

Do let me know your thoughts.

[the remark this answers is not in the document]
dirkkul: Sorry for the late reply, we're in the middle of renaming the company and that has kept me busy the last days.

I see the downsides of my approach, but I'd go for easier mantainability when there is a choice. I'm not happy with copying ~300 lines of code and I'm sure that one of them will be forgotten when there is the next addition/fix to the builder(s). 

As a compromise:
- You can move createFilterClause + createFieldsClause to builderBase and use them without problems from both implementations (do class.createFieldsClause() instead of mb.createFieldsClause(class)). That would cut the amount copied code by ~200 lines. 
- Moreover, we would need tests for each of the remaining duplicated functions + integration tests.

Does that work for you? @antas-marcin what do you think?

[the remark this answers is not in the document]
shreyanshshah27: I've made the suggested changes, let me know if anymore changes are required.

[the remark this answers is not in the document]
dirkkul: Cool, thanks!

I think an integration test would be great, as far as I see the "Do" function is also not used yet. And maybe an unittest with more than one property? Then we should have good test coverage and can merge afterwards :)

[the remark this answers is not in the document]
dirkkul: Would be great if you could rebase onto the latest main. I think I fixed the CI for PRs from forks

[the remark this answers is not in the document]
shreyanshshah27: I've re-based it with current main and also I've added few unit tests, let me know if this works or not after that will start working on adding integration test.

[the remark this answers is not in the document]
dirkkul: Tests run, thanks!

A few [seem to be failing](https://github.com/weaviate/weaviate-go-client/pull/106/commits/7a10e9c20c10bb6f95169b60cf8e0442b302acda), could you have a look? I can also have a look on Monday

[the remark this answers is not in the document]
shreyanshshah27: I just saw that, test : `TestMultiClassQueryBuilder/Multiple_filter#01`, is an error from my end, I'll fix that. But test : `TestMultiClassBM25Builder` looks like an pure ordering issue. Should we sort based on `className` while building the query?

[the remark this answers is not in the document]
dirkkul: > Should we sort based on className while building the query?

That should be pretty cheap, so yes let's do that!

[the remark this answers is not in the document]
shreyanshshah27: Hey, @dirkkul Sorry for the delay. Got busy with some priority work. I've fixed the unit test issue. While working on integration tests for `MultiClassBuilder`, got to know that these two `TestGraphQL_integration/Get_bm25_filter`, `TestGraphQL_integration/Get_hybrid_filter` integration tests cases are failing. It's the same way in main branch also. 

It would be great if you can quickly look into it and share your thoughts. Meanwhile I will add other integration tests for `MultiClassBuilder`.

v4/weaviate/graphql/multiclassbuilder.go
```diff
@@ -31,7 +32,13 @@ func (mb *MultiClassBuilder) Do(ctx context.Context) (*models.GraphQLResponse, e
 // build the GraphQL query string (not needed when Do is executed)
 func (mb *MultiClassBuilder) build() string {
 	var query string
-	for _, class := range mb.classBuilders {
+	// sorting classBuilder based on className to have consistent order in query
+	s := make(classBuildersSorter, 0, len(mb.classBuilders))
+	for _, k := range mb.classBuilders {
+		s = append(s, k)
+	}
+	sort.Sort(classBuildersSorter(s))
```
dirkkul, line 40: Instead of sorting  the classes, could we just add the map keys (eg the classnames to a slice) and sort those and then iterate over those afterwards? Then we do not need to add the sort interface.

Or another idea:
If we need a slice anyway, how about adding the keys directly to it? Then we would preserve the original order in which users add classes, which might be helpful?

v4/weaviate/graphql/multiclassbuilder.go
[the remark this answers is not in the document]
dirkkul: No worries, take your time :)

I added a comment, thanks for updating!

v4/weaviate/graphql/multiclassbuilder.go
shreyanshshah27: Makes sense to do that. I've also fixed the failing tests.

dirkkul: If Risotto is deleted again that should also fix those node tests that now error because of an additional class

shreyanshshah27: Sorry I didn't saw your comment before making `node test fix` commit. So the issue is, in most of the tests we are using `CreateTestSchemaAndData` function to setup classes and add data. Now, I didn't realize this before and because of that in each test i.e. node or backup we are creating one extra class `Risotto`. I thing it would be better to create a separate function just for `MultiClassBuilder` tests to setup and add data, which can insert `Risotto` class as well.

dirkkul (approved):

v4/weaviate/graphql/multiclassbuilder.go
shreyanshshah27: Yes there is no dependency. We can go ahead and merge this.

dirkkul: great, merging then. Thanks for your contribution!

I added two tickets with the [good first issue](https://github.com/weaviate/weaviate-go-client/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22) label in case you want to tackle something else.

PR #130, opened 2023-03-02
Fix readme imports

dirkkul:


tools/prepare_release.sh
```diff
@@ -27,7 +27,7 @@ if git rev-parse "$VERSION" >/dev/null 2>&1; then
 fi
 
 DIR=$(pwd -P)
-LIB_VER=$(basename $DIR)
+LIB_VER=v4
```
antas-marcin, line 30: can we get the version from parsing the go.mod file?

antas-marcin (approved):

Issue #112, opened 2023-01-19
Add wait_until_weaviate_is_ready function


dirkkul:
When starting it can take a few seconds until it is ready and the client can "use" it. Currently it's in the responsibility of the user to wait until the weavaite is ready. 

To improve the UX, we should let the client wait (with a timeout) until weaviate is ready and "v1/.well-known/ready" returns 404

Note: There is something comparable in the tests in `v4/weaviate/testenv/environment.go`, here is a PR for the python client that [implements something comparable](https://github.com/weaviate/weaviate-python-client/pull/217)

Sasmuin:
I will try this.

PR #135, opened 2023-03-09
Wait for weaviate

Sasmuin:
Closes https://github.com/weaviate/weaviate-go-client/issues/112. 

Implements the WaitForWeaviate function which automatically waits for weaviate to start before initializing the client. 

Due to the existing architecture of the client, there are two places where a first connection to weaviate happens:
1) When creating the config with authentication: Here the weaviate OIDC config is queried before the client is created.
2) After creating the client, before a user makes the first connection.

Due to this I added the WaitForWeaviate in two places:
1) In a new function AddAuthClient, that creates the correct http-client
2) In a member function of the client, eg client.WaitForWeavaite(time) that users can call after creating the client

### DIFF
--- a/test/misc/misc_test.go
+++ b/test/misc/misc_test.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/weaviate/weaviate-go-client/v4/test/testsuit"
+	"github.com/weaviate/weaviate-go-client/v4/weaviate"
 	"github.com/weaviate/weaviate-go-client/v4/weaviate/testenv"
 )
 
@@ -66,5 +67,33 @@ func TestMisc_integration(t *testing.T) {
 			fmt.Printf(err.Error())
 			t.Fail()
 		}
+	})
+}
+
+func TestMisc_connection_error(t *testing.T) {
+	t.Run("ready", func(t *testing.T) {
+		cfg := weaviate.Config{
+			Host:   "localhorst",
+			Scheme: "http",
+		}
+
+		client := weaviate.New(cfg)
+		isReady, err := client.Misc().ReadyChecker().Do(context.Background())
+
+		assert.NotNil(t, err)
+		assert.False(t, isReady)
+	})
+
+	t.Run("live", func(t *testing.T) {
+		cfg := weaviate.Config{
+			Host:   "localhorst",
+			Scheme: "http",
+		}
+
+		client := weaviate.New(cfg)
+		isReady, err := client.Misc().LiveChecker().Do(context.Background())
+
+		assert.NotNil(t, err)
+		assert.False(t, isReady)
 	})
 }
--- a/test/misc/misc_test.go
+++ b/test/misc/misc_test.go
@@ -7,7 +7,6 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/weaviate/weaviate-go-client/v4/test/testsuit"
-	"github.com/weaviate/weaviate-go-client/v4/weaviate"
 	"github.com/weaviate/weaviate-go-client/v4/weaviate/testenv"
 )
 
@@ -67,33 +66,5 @@ func TestMisc_integration(t *testing.T) {
 			fmt.Printf(err.Error())
 			t.Fail()
 		}
-	})
-}
-
-func TestMisc_connection_error(t *testing.T) {
-	t.Run("ready", func(t *testing.T) {
-		cfg := weaviate.Config{
-			Host:   "localhorst",
-			Scheme: "http",
-		}
-
-		client := weaviate.New(cfg)
-		isReady, err := client.Misc().ReadyChecker().Do(context.Background())
-
-		assert.NotNil(t, err)
-		assert.False(t, isReady)
-	})
-
-	t.Run("live", func(t *testing.T) {
-		cfg := weaviate.Config{
-			Host:   "localhorst",
-			Scheme: "http",
-		}
-
-		client := weaviate.New(cfg)
-		isReady, err := client.Misc().LiveChecker().Do(context.Background())
-
-		assert.NotNil(t, err)
-		assert.False(t, isReady)
 	})
 }
--- a/test/mock/wait_for_weaviate_test.go
+++ b/test/mock/wait_for_weaviate_test.go
@@ -5,6 +5,7 @@ import (
 	"net/http/httptest"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/weaviate/weaviate-go-client/v4/weaviate"
@@ -21,7 +22,7 @@ func TestWaitForWeaviate(t *testing.T) {
 
 	cfg := weaviate.Config{Host: strings.TrimPrefix(s.URL, "http://"), Scheme: "http"}
 	client := weaviate.New(cfg)
-	err := client.WaitForWeavaite(60)
+	err := client.WaitForWeavaite(60 * time.Second)
 	assert.Nil(t, err)
 }
 
@@ -33,6 +34,6 @@ func TestWaitForWeaviate_NoConnection(t *testing.T) {
 
 	cfg := weaviate.Config{Host: strings.TrimPrefix(s.URL, "http://"), Scheme: "http"}
 	client := weaviate.New(cfg)
-	err := client.WaitForWeavaite(5)
+	err := client.WaitForWeavaite(5 * time.Second)
 	assert.NotNil(t, err)
 }
--- a/test/mock/wait_for_weaviate_test.go
+++ b/test/mock/wait_for_weaviate_test.go
@@ -11,7 +11,7 @@ import (
 )
 
 func TestWaitForWeaviate(t *testing.T) {
-	// Returns the address of the auth server
+	// Tests the WaitForWeaviate function if a connection can be established
 	mux := http.NewServeMux()
 	mux.HandleFunc("/v1/.well-known/ready", func(w http.ResponseWriter, r *http.Request) {
 		w.Write([]byte(`{}`))
@@ -26,7 +26,7 @@ func TestWaitForWeaviate(t *testing.T) {
 }
 
 func TestWaitForWeaviate_NoConnection(t *testing.T) {
-	// Returns the address of the auth server
+	// Tests the WaitForWeaviate function if no connection can be established
 	mux := http.NewServeMux()
 	s := httptest.NewServer(mux)
 	defer s.Close()
--- a/test/mock/wait_for_weaviate_test.go
+++ b/test/mock/wait_for_weaviate_test.go
@@ -21,7 +21,7 @@ func TestWaitForWeaviate(t *testing.T) {
 
 	cfg := weaviate.Config{Host: strings.TrimPrefix(s.URL, "http://"), Scheme: "http"}
 	client := weaviate.New(cfg)
-	err := client.WaitForWeavaite(5)
+	err := client.WaitForWeavaite(60)
 	assert.Nil(t, err)
 }
 
--- a/test/mock/wait_for_weaviate_test.go
+++ b/test/mock/wait_for_weaviate_test.go
@@ -1,0 +1,38 @@
+package connection
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/weaviate/weaviate-go-client/v4/weaviate"
+)
+
+func TestWaitForWeaviate(t *testing.T) {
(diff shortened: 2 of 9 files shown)


test/mock/wait_for_weaviate_test.go
```diff
@@ -0,0 +1,38 @@
...
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/weaviate/weaviate-go-client/v4/weaviate"
+)
+
+func TestWaitForWeaviate(t *testing.T) {
+	// Returns the address of the auth server
```
dirkkul, line 14: wrong comment? (Same below)

weaviate/connection/rest.go
```diff
@@ -53,6 +54,38 @@ func NewConnection(scheme string, host string, httpClient *http.Client, headers
...
 
+// WaitForWeaviate waits until weaviate is started up and ready
+// expects weaviat at localhost:8080
+func (con *Connection) WaitForWeaviate(startup_period int) error {
+	if startup_period < 0 {
+		return errors.New("'startup_period' needs to be an integer larger than zero")
+	}
+
+	for i := 0; i < startup_period; i++ {
+		ctx, cancelFunc := context.WithTimeout(context.Background(), time.Second*3)
```
dirkkul, line 65: this should be one second, otherwise this would wait for up to 3 seconds ( I think)

weaviate/connection/rest.go
```diff
@@ -53,6 +54,38 @@ func NewConnection(scheme string, host string, httpClient *http.Client, headers
...
 }
 
+// WaitForWeaviate waits until weaviate is started up and ready
+// expects weaviat at localhost:8080
+func (con *Connection) WaitForWeaviate(startup_period int) error {
+	if startup_period < 0 {
+		return errors.New("'startup_period' needs to be an integer larger than zero")
+	}
+
+	for i := 0; i < startup_period; i++ {
```
dirkkul, line 64: You could use tickers instead of the loop here. Otherwise we would wait for the request timeout and then sleep afterwards. SO we would wait longer than the startup period. (Might also be a good test?)

Examples https://gobyexample.com/tickers

test/misc/misc_test.go
```diff
@@ -69,31 +68,3 @@ func TestMisc_integration(t *testing.T) {
 		}
 	})
 }
-
-func TestMisc_connection_error(t *testing.T) {
```
dirkkul, line 73: these tests could be readded?

Sasmuin: Does it make sense to readd this? I feel like this case is covered by WaitForWeaviate.

dirkkul: as long as we have the ready checker in the code base, we should have a test for it

dirkkul (approved):

weaviate/connection/rest.go
```diff
@@ -53,6 +54,39 @@ func NewConnection(scheme string, host string, httpClient *http.Client, headers
...
+		default:
+			isReady = false
+
+		}
+
+		cancelFunc()
+		if isReady {
+			return nil
+		}
+		if t.After(start_time.Add(time.Duration(startup_period) * time.Second)) {
```
parkerduckworth, line 83: Looking at this, I think it would be nicer to have this method's parameter type as a `time.Duration` rather than an int

weaviate/connection/rest.go
```diff
@@ -53,6 +54,39 @@ func NewConnection(scheme string, host string, httpClient *http.Client, headers
...
+			return fmt.Errorf("weaviate did not start up in %d seconds. Either the Weaviate URL %s is wrong or Weaviate did not start up in the interval given in 'startup_period'", startup_period, con.basePath)
```
parkerduckworth, line 84:
```
Weaviate URL %s
```

would recommend the `%q` formatter here

weaviate/connection/rest.go
```diff
@@ -53,6 +54,39 @@ func NewConnection(scheme string, host string, httpClient *http.Client, headers
...
+		}
+		fmt.Printf("Weaviate not yet up waiting another second. Current time: %v\n", t)
```
parkerduckworth, line 86: Probably a `log.Printf` would be better here, as a formatted timestamp would already be included. 

Also, this sentence is a bit hard to understand at first glance:
```
Weaviate not yet up waiting another second
```

weaviate/weaviate_client.go
```diff
@@ -48,6 +50,25 @@ func NewConfig(host string, scheme string, authConfig auth.Config, headers map[s
 	return &Config{Host: host, Scheme: scheme, Headers: headers, ConnectionClient: client}, nil
 }
 
+func AddAuthClient(cfg Config, authConfig auth.Config, startup_period int) (Config, error) {
```
parkerduckworth, line 53: As the above `startup_period` comment mentions, this parameter name/type could be improved

weaviate/weaviate_client.go
```diff
@@ -104,6 +125,11 @@ func New(config Config) *Client {
 	return client
 }
 
+// Waits for Weaviate to start.
+func (c *Client) WaitForWeavaite(startup_period int) error {
```
parkerduckworth, line 129: See above `startup_period` comments

parkerduckworth (changes requested):

Sasmuin:
I agree with the CLA.

weaviate/connection/rest.go
```diff
@@ -55,12 +55,12 @@ func NewConnection(scheme string, host string, httpClient *http.Client, headers
 }
 
 // WaitForWeaviate waits until weaviate is started up and ready
-func (con *Connection) WaitForWeaviate(startup_period int) error {
-	if startup_period < 0 {
-		return errors.New("'startup_period' needs to be an integer larger than zero")
+func (con *Connection) WaitForWeaviate(startupTimeout time.Duration) error {
+	if startupTimeout < 0 {
+		return errors.New("'startupTimeout' needs to be an integer larger than zero")
```
parkerduckworth, line 60: > `"'startupTimeout' needs to be an integer larger than zero"`

I don't think this check is necessary anymore, now that the param type is `time.Duration`.

weaviate/connection/rest.go
```diff
@@ -80,10 +80,10 @@ func (con *Connection) WaitForWeaviate(startup_period int) error {
 		if isReady {
 			return nil
 		}
-		if t.After(start_time.Add(time.Duration(startup_period) * time.Second)) {
-			return fmt.Errorf("weaviate did not start up in %d seconds. Either the Weaviate URL %s is wrong or Weaviate did not start up in the interval given in 'startup_period'", startup_period, con.basePath)
+		if t.After(startTime.Add(startupTimeout * time.Second)) {
```
parkerduckworth, line 83: > `t.After(startTime.Add(startupTimeout * time.Second))`

This should now be:
```go
t.After(startTime.Add(startupTimeout))
```

weaviate/weaviate_client.go
```diff
@@ -50,13 +51,13 @@ func NewConfig(host string, scheme string, authConfig auth.Config, headers map[s
 	return &Config{Host: host, Scheme: scheme, Headers: headers, ConnectionClient: client}, nil
 }
 
-func AddAuthClient(cfg Config, authConfig auth.Config, startup_period int) (Config, error) {
+func AddAuthClient(cfg Config, authConfig auth.Config, startupTimeout int) (Config, error) {
```
parkerduckworth, line 54: > startupTimeout int 

This should be `time.Duration`, no need to implicitly covert inside this function. It is good UX and idiomatic in Go to have a user just pass in an explicit `time.Duration` whenever setting timeouts, or dealing in elapsed time

weaviate/weaviate_client.go
```diff
@@ -50,13 +51,13 @@ func NewConfig(host string, scheme string, authConfig auth.Config, headers map[s
...
 	if authConfig == nil {
 		return cfg, nil
 	}
 
 	tmpCon := connection.NewConnection(cfg.Scheme, cfg.Host, nil, cfg.Headers)
-	err := tmpCon.WaitForWeaviate(startup_period)
+	err := tmpCon.WaitForWeaviate(time.Duration(startupTimeout))
```
parkerduckworth, line 60: See above comment about `time.Duration`

weaviate/weaviate_client.go
```diff
@@ -126,8 +127,8 @@ func New(config Config) *Client {
 }
 
 // Waits for Weaviate to start.
-func (c *Client) WaitForWeavaite(startup_period int) error {
-	return c.connection.WaitForWeaviate(startup_period)
+func (c *Client) WaitForWeavaite(startupTimeout int) error {
```
parkerduckworth, line 130: See above comment about `time.Duration`

weaviate/connection/rest.go
dirkkul: I think durations can be negative, so we would still need a check?

dirkkul: But the text needs to be adapted

weaviate/connection/rest.go
Sasmuin: I did change it now.

weaviate/connection/rest.go
parkerduckworth: Agreed, let's keep the check. I took a look at some of the `time` packages implementations, such as `NewTicker`, and there is a check for positive duration there as well.

weaviate/connection/rest.go
```diff
@@ -53,6 +54,39 @@ func NewConnection(scheme string, host string, httpClient *http.Client, headers
...
+			isReady = false
+
+		}
+
+		cancelFunc()
+		if isReady {
+			return nil
+		}
+		if t.After(startTime.Add(startupTimeout)) {
+			return fmt.Errorf("weaviate did not start up in %d seconds. Either the Weaviate URL %q is wrong or Weaviate did not start up in the interval given in 'startupTimeout'", startupTimeout, con.basePath)
```
parkerduckworth, line 84:
```go
fmt.Errorf("weaviate did not start up in %d seconds. Either the Weaviate URL %q is wrong or Weaviate did not start up in the interval given in 'startupTimeout'", startupTimeout, con.basePath)
```

Now that we are using `time.Duration` explicitly, we don't know for sure if the unit of time will necessarily be seconds. Luckily, `time.Duration` has a `String()` method, so this error message can now be:

```go
return fmt.Errorf("weaviate did not start up in %s. Either the Weaviate URL %q is wrong or Weaviate did not start up in the interval given in 'startupTimeout'", startupTimeout.String(), con.basePath)
```

parkerduckworth (changes requested):

weaviate/connection/rest.go
parkerduckworth: @Sasmuin yeah that's ok, this is idiotic Go

weaviate/connection/rest.go
```diff
@@ -53,6 +54,39 @@ func NewConnection(scheme string, host string, httpClient *http.Client, headers
 	return connection
 }
 
+// WaitForWeaviate waits until weaviate is started up and ready
+func (con *Connection) WaitForWeaviate(startupTimeout time.Duration) error {
+	if startupTimeout < 0 {
```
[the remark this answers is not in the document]
dirkkul: I think 0 would be "don't wait at all"?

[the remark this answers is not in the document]
parkerduckworth: Ok, if that is desired behavior, let's make that explicit. Also, the below error message should be changed, because it indicates that `0` is not a valid value

Sasmuin:
I also fixed that WaitForWeavaite always waited a second even if a connection could be established right away.

weaviate/connection/rest.go
```diff
@@ -56,34 +56,47 @@ func NewConnection(scheme string, host string, httpClient *http.Client, headers
...
 // WaitForWeaviate waits until weaviate is started up and ready
 func (con *Connection) WaitForWeaviate(startupTimeout time.Duration) error {
-	if startupTimeout < 0 {
-		return errors.New("'startupTimeout' needs to be a time.Duration larger than zero")
-	}
-	ticker := time.NewTicker(time.Second)
-	startTime := time.Now()
-	for {
-		t := <-ticker.C
+	switch {
```
dirkkul, line 59: total nitpick:
you don't need the switch here and can just do
```
if startupTimeout < 0{
	return errors.New("'startupTimeout' can not be smaller than zero")
} else if startupTimeout == 0{
	return nil
}
...rest of the code..
```

returning early for errors/easy caes and then having the "main path" at the base intendation reduces the cognitive load when reading code and is in general a good practice
([Keep code left](https://medium.com/@zacharyrsmith/keep-code-left-how-else-statements-are-hurting-your-code-39497e390b74)!)

Issue #125, opened 2023-02-20
Add method to check if a class exists


dirkkul:
Add a function that returns if a given class exists in the schema

Proposal signature (copied from python):
`client.schema.exists(className: str) ->bool`

Sasmuin:
I will do this.

PR #140, opened 2023-03-21
Add schema ClassExistenceChecker

Sasmuin:
Closes https://github.com/weaviate/weaviate-go-client/issues/125. Adds a ClassExistenceChecker that returns a boolean to show whether a class exists in a schema or not. 
The signature is: client.Schema().ClassChecker().WithClassName(className string).Do(ctx context.Context) -> bool, error.

### DIFF
--- a/test/schema/schema_test.go
+++ b/test/schema/schema_test.go
@@ -225,10 +225,12 @@ func TestSchema_integration(t *testing.T) {
 		err := client.Schema().ClassCreator().WithClass(schemaClass).Do(context.Background())
 		assert.Nil(t, err)
 
-		ok := client.Schema().ClassExistenceChecker().WithClassName("Band").Do(context.Background())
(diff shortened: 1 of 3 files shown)


test/schema/schema_test.go
```diff
@@ -215,6 +215,29 @@ func TestSchema_integration(t *testing.T) {
...
+		}
+
+		err := client.Schema().ClassCreator().WithClass(schemaClass).Do(context.Background())
+		assert.Nil(t, err)
+
+		ok, err := client.Schema().ClassExistenceChecker().WithClassName("Band").Do(context.Background())
+		assert.Nil(t, err)
+		assert.True(t, ok)
+
+		ok, err = client.Schema().ClassExistenceChecker().WithClassName("NonExistantClass").Do(context.Background())
```
dirkkul, line 232: NonExist**e**ntClass

Sasmuin: Oh right, I just copy pasted it from another test.

PR #141, opened 2023-03-23
Add api keys

Sasmuin:
Add the ability to use API Keys for authentification.

PR #145, opened 2023-03-27
Add api keys

dirkkul:
Superseeds https://github.com/weaviate/weaviate-go-client/pull/141

Add api keys.

Usage:

```
		cfg := weaviate.Config{Host: url, Scheme: "http", Headers: nil}
		var err error
		cfg, err = weaviate.AddAuthClient(cfg, authConf, 60*time.Second)
		assert.Nil(t, err)
		client := weaviate.New(cfg)
```


test/auth/auth_test.go
```diff
@@ -277,3 +277,60 @@ func getAccessToken(t *testing.T, weaviateUrl, user, pw string) (string, string)
 	require.Nil(t, err)
 	return tokens.AccessToken, tokens.RefreshToken
 }
+
+func TestApiKey(t *testing.T) {
```
antas-marcin, line 281: Here also test names are prefixed with `TestAuth_` so that one could also run only all tests from this file

weaviate/auth/auth.go
```diff
@@ -157,22 +157,33 @@ type BearerToken struct {
...
 		AccessToken: bt.AccessToken, RefreshToken: bt.RefreshToken, Expiry: time.Now().Add(time.Second * time.Duration(bt.ExpiresIn)),
 	})
 
-	return oauth2.NewClient(context.TODO(), tokenSource), nil
+	return oauth2.NewClient(context.TODO(), tokenSource), nil, nil
+}
+
+type ApiKeys struct {
+	ApiKey string
+}
```
antas-marcin, line 182: Wouldn't it be better to call this struct just `ApiKey` like that:
```
type ApiKey struct {
  Value string
}
```

.github/workflows/tests.yaml
```diff
@@ -44,8 +38,59 @@ jobs:
...
+        with:
+          username: ${{secrets.DOCKER_USERNAME}}
+          password: ${{secrets.DOCKER_PASSWORD}}
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.19
+          cache: true
+      - name: Run tests
+        run: ( for pkg in $(go list ./... | grep 'weaviate-go-client/v4/test/' | grep -v 'auth'); do if ! go test -v -count 1 -race "$pkg"; then echo "Test for $pkg failed" >&2; false; exit; fi; done)
```
antas-marcin, line 69: Maybe we could wrap this command into `./tools/run_tests.sh` script?

.github/workflows/tests.yaml
```diff
@@ -44,8 +38,59 @@ jobs:
...
+  unit-tests:
+    name: Unit tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.19
+          cache: true
```
antas-marcin, line 79: This part is not added to `tests-deprecated` step, is that intensional?

.github/workflows/tests.yaml
```diff
@@ -44,8 +38,59 @@ jobs:
...
+      - uses: actions/checkout@v3
+      - name: Login to Docker Hub
+        uses: docker/login-action@v2
+        if: ${{ !github.event.pull_request.head.repo.fork }}
+        with:
+          username: ${{secrets.DOCKER_USERNAME}}
+          password: ${{secrets.DOCKER_PASSWORD}}
+      - name: Run tests
+        run:
+          go test -v -count 1 -race test_deprecated/*.go
```
antas-marcin, line 96: Actually if we would have `./tools/run_tests.sh` script then we could add modifiers to them for all tests, so that for example this part would be invoked with `./tools/run_tests.sh --deprecated`

dirkkul: done, please have a look :)

.github/workflows/tests.yaml
dirkkul: added

tools/run_tests.sh
```diff
@@ -0,0 +1,73 @@
...
+  then
+    echo_green "Run all deprecated tests..."
+    go test -v -count 1 -race test_deprecated/*.go
+    echo_green "deprecated tests successful"
+  fi
+  echo "Done!"
+}
+
+function run_integration_tests() {
+    for pkg in $(go list ./... | grep 'weaviate-go-client/v4/test/' | grep -v 'auth' ); do
```
antas-marcin, line 57: You can even make this part so that with every new release of the client we would not have to update the package version:
```
    GO_MOD_PKG=$(head -n 1 go.mod | sed -r 's/module //g')
    LIB_VER=$(basename $GO_MOD_PKG)
    PKG_TEST="weaviate-go-client/$LIB_VER/test/"
    for pkg in $(go list ./... | grep $PKG_TEST | grep -v 'auth' ); do
```

tools/run_tests.sh
```diff
@@ -0,0 +1,73 @@
+function main() {
```
antas-marcin, line 1: Please add shebang intepreter directive at the very top of this script:
```
#!/usr/bin/env bash
```

test/auth/auth_test.go
```diff
@@ -277,3 +266,83 @@ func getAccessToken(t *testing.T, weaviateUrl, user, pw string) (string, string)
...
+				client = weaviate.New(*cfg)
+
+			} else {
+				cfg := weaviate.Config{Host: url, Scheme: "http", StartupTimeout: 60 * time.Second, AuthConfig: authConf, Headers: tc.headers}
+				var err error
+				client, err = weaviate.NewClient(cfg)
+				assert.Nil(t, err)
+			}
+
+			AuthErr := client.Schema().AllDeleter().Do(context.TODO())
```
antas-marcin, line 300: I think this variable can start with a small letter

test/auth/auth_test.go
```diff
@@ -277,3 +266,83 @@ func getAccessToken(t *testing.T, weaviateUrl, user, pw string) (string, string)
...
+	}{
+		{headers: nil}, {headers: map[string]string{}}, {headers: map[string]string{"Key": "Val"}},
+	}
+	for _, tc := range tests {
+		t.Run("headers", func(t *testing.T) {
+			cfg := weaviate.Config{Host: url, Scheme: "http", StartupTimeout: 60 * time.Second, AuthConfig: auth.ApiKey{Value: "wrong_key"}, Headers: tc.headers}
+			client, err := weaviate.NewClient(cfg)
+			assert.Nil(t, err)
+
+			AuthErr := client.Schema().AllDeleter().Do(context.TODO())
```
antas-marcin, line 319: here also, can be `authErr`

antas-marcin (approved):

Issue #146, opened 2023-04-07
improve getting started ergonomics


tmc:
Currently, if an improperly passed Host in a weaviate.Config is passed the returned error is:

`status code: -1, error: check the DerivedFromError field for more information`

It's a bit atypical in go to not wrap the underlying error.

I had to do some hoop jumping like:
```
	meta, err := client.Misc().MetaGetter().Do(context.Background())
	if err != nil {
		wce := &fault.WeaviateClientError{}
		if errors.As(err, &wce) {
			return fmt.Errorf("error with meta retrieval: %w", wce.DerivedFromError)
		}
	}
```

To detect an underlying `Get "https://https//my-sandbox-cluster-xxxx.weaviate.network/v1/meta": dial tcp: lookup https: no such host`.

I recommend you wrap your underlying errors in the default Error() implementation in your custom error type to make this more straightforward.

tmc:
Similarly, your `*ErrorResponse` type should likely implement the error interface.

PR #147, opened 2023-04-07
Fix WeaviateClientError's Error method to return proper error message

antas-marcin:
Addresses #146 issue

### DIFF
--- a/test/misc/misc_test.go
+++ b/test/misc/misc_test.go
@@ -74,6 +74,7 @@ func TestMisc_connection_error(t *testing.T) {
 
 		assert.NotNil(t, err)
 		assert.False(t, isReady)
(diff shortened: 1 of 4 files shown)


tmc (changes requested):
This seems like more indirection than necessary, I'd instead just expect that if DerivedFromError is not nil that it's included in the error string.

This could be as simple as:

```go
msg := uce.Msg
if uce.DerivedFromError != nil {
    msg := fmt.Sprintf("%s %v", uce.Msg, uce.DerivedFromError)
}
return fmt.Sprintf("status code: %v, error: %v", uce.StatusCode, msg)
}

aliszka (approved):

dirkkul (approved):

PR #154, opened 2023-05-01
Token refresh

Sasmuin:
Closes https://github.com/weaviate/weaviate-go-client/issues/137. Allows users with an unstable connection to try refreshing the authentication token several times in order to avoid unauthenticated clients.

### DIFF
--- a/test/auth/auth_mock_test.go
+++ b/test/auth/auth_mock_test.go
@@ -209,13 +209,21 @@ func TestAuthMock_RefreshUserPWAndToken(t *testing.T) {
 }
 
 func TestAuthMock_RefreshTokenTimeout(t *testing.T) {
+	tokenRefreshTime := time.Now()
 	expirationTimeToken := uint(31)
 	expirationTimeRefreshToken := 2
 	authConfig := auth.BearerToken{AccessToken: AccessToken, ExpiresIn: expirationTimeToken, RefreshToken: RefreshToken}
+	firstTime := true
+	pointerFirstTime := &firstTime
 
 	// endpoint for access tokens
 	muxToken := http.NewServeMux()
-	muxToken.HandleFunc("/meta", func(w http.ResponseWriter, r *http.Request) {
+	muxToken.HandleFunc("/auth", func(w http.ResponseWriter, r *http.Request) {
+		if *pointerFirstTime {
+			time.Sleep(6 * time.Second)
+		}
+		*pointerFirstTime = false
+		tokenRefreshTime = time.Now() // update time when the tokens where refreshed the last time
 		w.Header().Set("Content-Type", "application/json")
 		w.Write([]byte(
 			fmt.Sprintf(`{"access_token": "%v", "expires_in": %v, "refresh_token": "%v", "refresh_expires_in" :  %v}`,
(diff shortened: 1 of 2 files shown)


test/auth/auth_mock_test.go
```diff
@@ -158,13 +158,14 @@ func TestAuthMock_RefreshUserPWAndToken(t *testing.T) {
 			muxToken := http.NewServeMux()
 			muxToken.HandleFunc("/auth", func(w http.ResponseWriter, r *http.Request) {
 				// refresh token cannot be expired
-				assert.True(t, time.Now().Sub(tokenRefreshTime).Seconds() < float64(expirationTimeRefreshToken))
+				log.Printf("%v", time.Since(tokenRefreshTime).Seconds())
```
dirkkul, line 161: leftover from debugging?

test/auth/auth_mock_test.go
```diff
@@ -158,13 +158,14 @@ func TestAuthMock_RefreshUserPWAndToken(t *testing.T) {
...
+				assert.True(t, time.Since(tokenRefreshTime).Seconds() < float64(expirationTimeRefreshToken))
 
 				tokenRefreshTime = time.Now() // update time when the tokens where refreshed the last time
 				w.Header().Set("Content-Type", "application/json")
 				w.Write([]byte(
 					fmt.Sprintf(`{"access_token": "%v", "expires_in": %v, "refresh_token": "%v", "refresh_expires_in" :  %v}`,
-						AccessToken, expirationTimeToken, RefreshToken, expirationTimeRefreshToken)))
+						AccessToken, uint(12), RefreshToken, expirationTimeRefreshToken)))
```
dirkkul, line 168: leftover from debugging?

weaviate/connection/rest.go
```diff
@@ -107,28 +107,37 @@ func (con *Connection) startRefreshGoroutine(transport *oauth2.Transport) {
 		log.Printf("Error during token refresh, getting token: %v", err)
 		return
 	}
+
+	if time.Until(token.Expiry) < 0 {
```
dirkkul, line 111: the refresh token might still be valid?

weaviate/connection/rest.go
```diff
@@ -107,28 +107,37 @@ func (con *Connection) startRefreshGoroutine(transport *oauth2.Transport) {
...
+				timeToSleep := time.Until(token.Expiry) - time.Second*10
+				if timeToSleep > 0 {
+					time.Sleep(timeToSleep)
+				}
+				token, err = transport.Source.Token()
 				if err != nil {
-					log.Printf("Error during token refresh, rest request: %v", err)
+					log.Printf("Error during token refresh, getting token: %v", err)
+				}
+				if time.Until(token.Expiry) < 0 {
```
dirkkul, line 136: as long as the refresh token is still valid we can continue trying to refresh

weaviate/connection/rest.go
```diff
@@ -107,28 +107,37 @@ func (con *Connection) startRefreshGoroutine(transport *oauth2.Transport) {
...
+					log.Printf("Requested token is expired. Stop requesting new access token.")
 					return
 				}
+				time.Sleep(time.Second)
```
dirkkul, line 140: I think this could have a better structure:
```
*expiration check*
token, err = transport.Source.Token()
if err != nil {
    log.Printf("Error during token refresh, getting token: %v", err)
    time.sleep(time.Second)
}else{
      timeToSleep := time.Until(token.Expiry) - time.Second*10
     if timeToSleep > 0 {
	time.Sleep(timeToSleep)
   }
}
```

test/auth/auth_mock_test.go
Sasmuin: This was actually done on purpose to check if the sleeping works correctly

weaviate/connection/rest.go
Sasmuin: Sadly here is no easy way to access the validity of the refresh token

test/auth/auth_mock_test.go
dirkkul: could you add a comment explaining why the 12 is there?

Sasmuin: Done

dirkkul (approved):

PR #157, opened 2023-05-04
Fix tests for 1.19

donomii:
It appears some unit tests require a running weaviate, this starts the containers before the unit tests.

Also fixes the tests changed by the string->text patch

### DIFF
--- a/test/schema/schema_test.go
+++ b/test/schema/schema_test.go
@@ -2,7 +2,6 @@ package schema
 
 import (
 	"context"
-	"os"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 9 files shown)


test/schema/schema_test.go
```diff
@@ -508,15 +509,28 @@ func TestSchema_integration(t *testing.T) {
...
-	t.Run("Create class with explicit replication config", func(t *testing.T) {
+	t.Run("tear down weaviate", func(t *testing.T) {
+		err := testenv.TearDownLocalWeaviate()
+		if err != nil {
+			t.Fatalf("failed to tear down weaviate: %s", err)
+		}
+	})
+}
+
+/*
```
antas-marcin, line 520: Why is this test commented?

donomii: Because it was constantly failing, and I'm not convinced that the go client is the right place for doing cluster testing.

donomii: Not when combined with basic functionality tests for the client

donomii: We can put it in its own file later

antas-marcin: maybe instead of commenting it, please add `t.Skip()` statement, it will be then easy to find

antas-marcin (approved):

PR #178, opened 2023-08-13
ContainsAny & ContainsAll Filter Operators

antas-marcin:


test/graphql/graphql_test.go
```diff
@@ -1946,6 +1948,417 @@ func TestGraphQL_MultiTenancy(t *testing.T) {
...
+			}
+
+			t.Run("create class", func(t *testing.T) {
+				cleanup()
+				class := &models.Class{
+					Class: className,
+					Properties: []*models.Property{
+						{
+							Name:     "color",
+							DataType: []string{schema.DataTypeText.String()},
```
aliszka, line 1971: you can use `schema.DataTypeText.PropString()` instead of `[]string{schema.DataTypeText.String()}`
(also applies to other data types)

aliszka (approved):

PR #192, opened 2023-11-13
Add support for cross references in gRPC Batch API

antas-marcin:
This PR adds support for `cross-references` in `gRPC Batch API` calls.


tsmith023 (approved):
As always, I don't like how the Go client is coupled to the server-side auto-generated swagger structs like `models.Object` but this is the convention now so it's good to stick to it!

PR #199, opened 2024-01-09
init backup compression config

moogacs:


moogacs:
should be revisited later on

PR #222, opened 2024-07-08
Add support for multi target vector search

dirkkul:


test/graphql/multi_target_near_vector_test.go
```diff
@@ -0,0 +1,65 @@
...
+		var inner = []struct {
+			nvi *graphql.NearMultiVectorArgumentBuilder
+		}{
+			{to.nvo.WithVector([]float32{1, 0, 0})},
+			{to.nvo.WithVectorPerTarget(map[string][]float32{"first": {1, 0}, "second": {1, 0, 0}})},
+		}
+		for _, ti := range inner {
+			resp, err := client.GraphQL().Get().WithNearMultiVector(ti.nvi).WithClassName(class.Class).Do(ctx)
+			require.Nil(t, err)
+			require.NotNil(t, resp)
```
dirkkul, line 62: do you want to assert some of the returns?

weaviate/graphql/nearmultivectorbuilder.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func (m *NearMultiVectorArgumentBuilder) WithVectorPerTarget(vectorPerTarget map[string][]float32) *NearMultiVectorArgumentBuilder {
+	if len(vectorPerTarget) > 0 {
+		m.vectorPerTarget = vectorPerTarget
+	}
+	return m
+}
+
+func (m *NearMultiVectorArgumentBuilder) WithVector(vector []float32) *NearMultiVectorArgumentBuilder {
```
dirkkul, line 102: did you remove the "two-stage" builder on purpose? :) 

I added the them so that users at first only have the choice of choosing a combination type and then HAVE to chose a vector-type. Was more playing around and am not sure if that was a good idea, but would like to discuss it :)

weaviate/graphql/nearmultivectorbuilder.go
```diff
@@ -0,0 +1,147 @@
...
+	if len(m.targetVectors) > 0 {
+		targets := m.toTargets()
+		weights, err := json.Marshal(targets.weights)
+		if err != nil {
+			panic(fmt.Errorf("failed to unmarshal near multi vector search weights: %s", err))
+		}
+		clause = append(clause, fmt.Sprintf("targets:{combinationMethod: %s, targetVectors: %s, weights: %s}", targets.combinationMethod, targets.targetVectors, string(weights)))
+	}
+	return fmt.Sprintf("nearVector:{%v}", strings.Join(clause, " "))
+}
```
dirkkul, line 147: can you add some unit tests?

weaviate/graphql/nearmultivectorbuilder.go
tsmith023: Yes, I removed it so that users have the option of not specifying the combination type thereby resorting to the default. This was to bring it in-line with the other clients, e.g. py-v4 doesn't require a method, right?
However, I'm not sold on it, do you think it would be useful to enforce a choice from the user?

test/graphql/multi_target_near_vector_test.go
tsmith023: I thought it easier to assume the functionality of Weaviate is correct here by only asserting that the query succeeded. I think that assumption is valid provided that I add unit tests to the `build()` method, which I'll do below, since we're testing that the Go code correctly creates a valid GQL query with the behaviour of the query tested within Weaviate itself. WDYT?

dirkkul: I think at least one assertion would be great, just to be sure

weaviate/graphql/nearmultivectorbuilder.go
dirkkul: >  However, I'm not sold on it, do you think it would be useful to enforce a choice from the user?

Would be more about guiding users through the choices. They HAVE to add a vector (either single vector or vector per target), but you are right, the combination method has a default.

Let's leave it like it is

weaviate/graphql/nearmultivectorbuilder.go
tsmith023: Unit tests of `.build()` are flakey because of unordered behaviour of keys in `vectorPerTarget` and `targetCombination.Weights` so can't guarantee the same string each time. Can you see an easy way to avoid this? As I cannot :(

test/graphql/multi_target_near_vector_test.go
tsmith023: Ah, I thought my `.NotNil` assertion was enough but failed to factor in the GQL query returning errors in a successful HTTP request. Have added assertions that highlighted bugs, thanks!

weaviate/graphql/nearmultivectorbuilder.go
dirkkul: you could work with strings.contains() to match substrings?

dirkkul (commented):
LGTM

PR #225, opened 2024-07-11
Fix hanging auth tests by making them aware of differing ports

tsmith023:


test/misc/misc_version_check_test.go
```diff
@@ -94,16 +96,13 @@ func TestMisc_empty_version_check(t *testing.T) {
...
-		Host:    "localhost:8080",
-		Scheme:  "http",
-		Headers: map[string]string{},
-	}
-
-	client := weaviate.New(*cfg)
-
 	t.Run("Weaviate is not live, perform live check", func(t *testing.T) {
+		cfg := &weaviate.Config{
+			Host:    "localhost:8080",
```
dirkkul, line 101: maybe use another port than 8080 - we have a constant "NoWeaviatePort" or something like that somewhere

tsmith023: Ahh, this should actually be the port returned by `GetPortAndAuthPw` so thanks for the call-out

Issue #224, opened 2024-07-11
Add timeout to grpc dial


dirkkul:
We do not set a timeout here https://github.com/weaviate/weaviate-go-client/blob/main/weaviate/connection/grpc.go#L68 - in case of a missconfiguration we will hang.

PR #226, opened 2024-07-11
Add grpc client init timeout with tests validating usage

tsmith023:
Closes https://github.com/weaviate/weaviate-go-client/issues/224

### DIFF
--- a/test/client/startup_test.go
+++ b/test/client/startup_test.go
@@ -1,0 +1,47 @@
+package client
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	"github.com/weaviate/weaviate-go-client/v4/test/testsuit"
+	"github.com/weaviate/weaviate-go-client/v4/weaviate"
+	"github.com/weaviate/weaviate-go-client/v4/weaviate/auth"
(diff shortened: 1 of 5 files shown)


test/client/startup_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+func TestStartupTimeout_GRPC(t *testing.T) {
+	port, _, authEnabled := testsuit.GetPortAndAuthPw()
+	cfg := weaviate.Config{
+		Host:   fmt.Sprintf("localhost:%v", port),
+		Scheme: "http",
+		GrpcConfig: &grpc.Config{
+			Host: fmt.Sprintf("localhost:%v", testsuit.NoWeaviateGRPCPort),
+		},
+		StartupTimeout: 1000 * time.Millisecond,
```
dirkkul, line 41: or one second? 😆

dirkkul (approved):

test/client/startup_test.go
tsmith023: My IDE did not prompt that as an option on `time`! Will change 😁

stevenh:
This is actually an API change which should have caused a major version bump.

It's easy to fix but for to abide by go package sematic versioning this should be kept this in mind in the future.

PR #228, opened 2024-07-18
Add `MultiTargetArgumentBuilder` for use in `WithTargets` methods

tsmith023:


dirkkul (approved):

weaviate/graphql/nearvectorbuilder.go
```diff
@@ -59,7 +67,10 @@ func (b *NearVectorArgumentBuilder) build() string {
 		}
 		clause = append(clause, fmt.Sprintf("vector: %s", string(vectorB)))
 	}
-	if len(b.targetVectors) > 0 {
+	if b.targets != nil {
+		clause = append(clause, fmt.Sprintf("targets:{%s}", b.targets.build()))
+	}
+	if len(b.targetVectors) > 0 && b.targets == nil {
```
antas-marcin, line 73: I think that you are missing `vectorPerTarget` field to make `targets` working in `nearVector`

tsmith023: In that case, users should use the near multi target vector builder. Although, now that you mention it, perhaps it should also just be handled within the near vector builder rather than splitting them up 🤔

antas-marcin: hmm wouldn't it feel more coherent to have `vectorPerTarget` field in here and remove `multi target vector builder`? bc for all other arguments you are setting the targets within them but only this one gets divided into 2 different arguments...

tsmith023: Yeh, I agree now looking at it again wrt the other clients we've developed. I'll open a PR into this one refactoring the support for `nearVector`

tsmith023: @antas-marcin: https://github.com/weaviate/weaviate-go-client/pull/232

PR #233, opened 2024-07-25
Add journey test for batch inserting objects

tsmith023:


tsmith023:
Test takes too long (~5m)

test/stress/mt_batching_test.go
```diff
@@ -0,0 +1,119 @@
...
+	className := "Collection"
+	client := testsuit.CreateTestClient(true)
+	ctx := context.Background()
+
+	defer func() {
+		err := client.Schema().ClassDeleter().WithClassName(className).Do(ctx)
+		require.Nil(t, err)
+	}()
+
+	t.Run("Create multi-tenancy collection", func(t *testing.T) {
```
dirkkul, line 25: can we have this outside of a t.Run() ?
Otherwise you will aways have to run the complete test and cannot run the individual subtest on their own

test/stress/mt_batching_test.go
```diff
@@ -0,0 +1,119 @@
...
+					WithPath([]string{"price"}).WithValueNumber(float64(100))).
+				WithTenant(tenant).
+				WithFields(graphql.Field{Name: "name"}).
+				Do(ctx)
+			if err != nil {
+				t.Fatalf("Search objects usage error: %v", err)
+			}
+			if res.Errors != nil {
+				t.Fatalf("Search objects GQL error: %+v", res.Errors[0])
+			}
```
dirkkul, line 114: just use `require.nil(t, err)` ?

test/stress/mt_batching_test.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/weaviate/weaviate/entities/models"
+)
+
+func TestMTBatching_stress(t *testing.T) {
+	className := "Collection"
+	client := testsuit.CreateTestClient(true)
+	ctx := context.Background()
+
+	defer func() {
+		err := client.Schema().ClassDeleter().WithClassName(className).Do(ctx)
```
dirkkul, line 21: I would also run this before creating the class, so any leftovers are deleted

test/stress/mt_batching_test.go
```diff
@@ -0,0 +1,119 @@
...
+		}).Do(ctx)
+		if err != nil {
+			t.Fatalf("Create collection: %v", err)
+		}
+	})
+
+	vector := make([]float32, 16)
+	for i := range vector {
+		vector[i] = float32(i) / 16
+	}
```
dirkkul, line 55: not really an issue in this case, but FYI: If you always use the same vector inserts and searches are very slow

test/stress/mt_batching_test.go
```diff
@@ -0,0 +1,115 @@
...
+		MultiTenancyConfig: &models.MultiTenancyConfig{
+			AutoTenantActivation: true,
+			AutoTenantCreation:   true,
+			Enabled:              true,
+		},
+		Vectorizer: "none",
+	}).Do(ctx)
+	if err != nil {
+		t.Fatalf("Create collection: %v", err)
+	}
```
dirkkul, line 50: require?

test/stress/mt_batching_test.go
```diff
@@ -0,0 +1,115 @@
...
+				WithClassName(className).
+				WithWhere(where.WithOperator(filters.LessThan).
+					WithPath([]string{"price"}).WithValueNumber(float64(100))).
+				WithTenant(tenant).
+				WithFields(graphql.Field{Name: "name"}).
+				Do(ctx)
+			require.Nil(t, err)
+			if res.Errors != nil {
+				t.Fatalf("Search objects GQL error: %+v", res.Errors[0])
+			}
```
dirkkul, line 110: also require.Length?

test/stress/mt_batching_test.go
tsmith023: Ah, really? Makes sense really and I'll change anyway

dirkkul: you will have a ton of warnings in your weaviate log if you do this😆

dirkkul (approved):

PR #239, opened 2024-08-19
Add support for multiple vectors per target

dirkkul:


weaviate/graphql/multitargetbuilder.go
```diff
@@ -9,16 +9,17 @@ import (
...
 func (m *MultiTargetArgumentBuilder) getCombinationMethod() string {
 	combinationMethod := ""
 	if m.targetCombination == nil {
 		return combinationMethod
 	}
-	switch m.targetCombination.Type {
+	switch *m.targetCombination {
 	case dto.Sum:
 		combinationMethod = "sum"
 	case dto.Average:
```
bevzzz, line 25: Nit: because `TargetCombinationType` is its own type you could move the `switch` to statement into its `String() string` method. 

```go
func (t TargetCombinationType) String() string {
  switch t {
    case Sum:
      return "sum"
   // ...
   }
}
```

Assuming they have constant mapping, i.e. `Sum` should always translate to `"sum"` and so on.

weaviate/graphql/nearvectorbuilder_test.go
```diff
@@ -79,9 +79,32 @@ func TestNearMultiVectorBuilder_build(t *testing.T) {
...
+		require.NotContains(t, str, "combinationMethod:")
+		require.NotContains(t, str, "weights:")
+	})
+
+	t.Run("No combination with multiple vectors per target", func(t *testing.T) {
+		vector := NearVectorArgumentBuilder{}
+		str := vector.WithVectorsPerTarget(map[string][][]float32{"one": {{1, 2, 3}, {7, 8, 9}}, "two": {{4, 5, 6}}}).build()
+		require.Contains(t, str, "vectorPerTarget: ")
+		require.Contains(t, str, "one: [[1,2,3],[7,8,9]]")
+		require.Contains(t, str, "two: [4,5,6]")
```
bevzzz, line 92: Will the server be able to handle `[[4,5,6]]` correctly? Like see that it's just 1 vector and not do any extra work.
Always sending `[][]float32` could simplify the code by removing [this special case](https://github.com/weaviate/weaviate-go-client/pull/239/files#diff-f350dc094391aa08961984075cbc93460640cdcbdaeeacdb538dfd5dbc7e4154R87-R98).

test/graphql/multi_target_search_test.go
```diff
@@ -208,3 +209,58 @@ func TestMultiTargetNearVector(t *testing.T) {
...
+	require.Nil(t, client.Schema().ClassCreator().WithClass(class).Do(ctx))
+	defer client.Schema().ClassDeleter().WithClassName(class.Class).Do(ctx)
+
+	objs := []*models.Object{
+		{Class: class.Class, Vectors: map[string]models.Vector{"first": []float32{1, 0, 0}, "second": []float32{1, 0, 0}}, ID: strfmt.UUID(uuid.New().String())},
+		{Class: class.Class, Vectors: map[string]models.Vector{"first": []float32{0, 0, 1}, "second": []float32{0, 0, 1}}, ID: strfmt.UUID(uuid.New().String())},
+	}
+
+	_, err = client.Batch().ObjectsBatcher().WithObjects(objs...).Do(ctx)
+	require.Nil(t, err)
```
bevzzz, line 238:
```suggestion
	require.NoError(t, err)
```

This would give a more readable message, like "Received unexpected error  'Whaam!'" instead of "expected nil, got &Error{...}"

dirkkul: ohh, nice one!

weaviate/graphql/nearvectorbuilder_test.go
dirkkul: yep, works 👍  Removed the extra handling

weaviate/graphql/get.go
```diff
@@ -228,6 +228,11 @@ func (gb *GetBuilder) Do(ctx context.Context) (*models.GraphQLResponse, error) {
 	return runGraphQLQuery(ctx, gb.connection, gb.build())
 }
 
+// Do execute the GraphQL query
```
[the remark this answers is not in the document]
dirkkul: changed to "Build ..."

weaviate/graphql/multitargetbuilder.go
dirkkul: that is imported from weaviate and we do not need it there. In general I like the proposal, but I think in this case we should leave it like it is

bevzzz: Agreed 👍 

---

Btw, if we still wanted to have `String()` method handle it without cluttering `weaviate` core, we could alias the type:

```go
// In weaviate-go-client
import "github.com/weaviate/weaviate"

type TargetCombinationType weaviate.TargetCombinationType
func (t TargetCombinationType) String() string { ... }
```

bevzzz:
LGTM 🍀

PR #244, opened 2024-10-12
gh-240: support backup cancellation

### LINKED ISSUE #240: add Backup cancellation from client
there are situations were backup gets stuck and in those situations we would like to cancel the created (stuck) backups. 

so Weaviate will support backup cancellation from `1.27+` and in order for clients users to be able to cancel  backups

[that is the new specification for deleting a backup ](https://github.com/weaviate/weaviate/pull/5639/files#diff-6e46d1a078f97315e53e6f21a97a0d18ac955212f3272ad75cfe0db74a248398R4862)

Screenshot 2024-08-22 at 16 41 26

bevzzz:
Closes #240 

## What is changed?

Extended client to support "cancel backup" functionality.
Bumped core weaviate/weaviate lib version to reference new backup state and, subsequently, the packages Go version (1.21 -> 1.22)

## How is it tested?
- New integration test in `TestBackups_integration`


go.mod
```diff
@@ -1,27 +1,28 @@
 module github.com/weaviate/weaviate-go-client/v4
 
-go 1.21
+go 1.22
+
+toolchain go1.22.8
```
dirkkul, line 5: what does this do?

test/backup/backup_test.go
```diff
@@ -901,6 +903,35 @@ func TestBackups_integration(t *testing.T) {
...
+	t.Run("cancel backup", func(t *testing.T) {
+		testsuit.CreateTestSchemaAndData(t, client)
+		defer testsuit.CleanUpWeaviate(t, client)
+
+		class := "Pizza"
+		backend := backup.BACKEND_FILESYSTEM
+		id := fmt.Sprint(random.Int63())
+		ctx := context.Background()
+
+		// Arrange: Check fixture data exists and create a backup of it
```
dirkkul, line 916: am not the biggest fan of these standard comments 😅  - for me they make sense for bigger tests but create noise for these small ones

test/schema/default_config_test.go
```diff
@@ -73,7 +73,8 @@ var defaultVectorIndexConfig = map[string]interface{}{
 }
 
 var defaultReplicationConfig = &models.ReplicationConfig{
-	Factor: 1,
+	Factor:                           1,
+	ObjectDeletionConflictResolution: models.ReplicationConfigObjectDeletionConflictResolutionPermanentDeletion,
```
dirkkul, line 77: this was merged prematurely and we will most likely change the name again (ObjectDeletionConflictResolution) - do we need to set it in this PR?

go.mod
bevzzz: `toolchain` specifies the version of [Go Toolchain](https://www.google.com/search?client=safari&rls=en&q=go+toolchain&ie=UTF-8&oe=UTF-8) (things that you can use with `go ...` command. Usually its version is the same as the one in `go` directive, but they may differ for some versions (see **Go versions** in the link) and then we need an explicit `toolchain` directive.

I'll just set `go 1.22.0` so that the versions match and remove `toolchain`.

test/schema/default_config_test.go
bevzzz: Without this field [one of the integration tests is failing](https://github.com/weaviate/weaviate-go-client/actions/runs/11313124509/job/31461500050#step:5:1668). So it's either setting the field or t.Skip-ing the test.

I'd do the former -- once that field is renamed, the next person to do a PR to this repo will be reminded to update it by the compiler.

weaviate/backup/backup.go
```diff
@@ -48,3 +48,10 @@ func (s *API) RestoreStatusGetter() *BackupRestoreStatusGetter {
 		connection: s.connection,
 	}
 }
+
+// Canceller creates a builder for "cancel backup" request.
+func (s *API) Canceller() *BackupCanceller {
```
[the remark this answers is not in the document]
tsmith023: Canceller = 🇬🇧 
Canceler = 🇺🇸 

So take your pick! If we have a standard then we should stick to it but I don't know whether we have such a thing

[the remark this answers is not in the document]
bevzzz: I'm going to venture a guess that Tommy may prefer the double-L spelling (according to Google) 😄:

image

Here though, I'm going to side with Chat and use the single-L spelling to highlight the `Cancel() -> Canceler` "grammar" (verb+"er")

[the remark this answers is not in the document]
bevzzz: > Canceller = 🇬🇧
Canceler = 🇺🇸

Oh, I was but a second too late to foretell that!

go.mod
bevzzz: From the article:

> The standard Go toolchains are named goV where V is a Go version denoting a beta release, release candidate, or release. For example, go1.21rc1 and **go1.21.0 are toolchain names**; **go1.21 and go1.22 are not** (the initial releases are go1.21.0 and go1.22.0), but **go1.20 and go1.19 are**.

It's a bit tricky: `go` and `toolchain` directives "understand" versions differently. I'm guessing that before (with go1.20 and go1.19) you could just specify the major one.
But for "toolchain" go1.22 is not a valid toolchain name.

I just stumbled upon this recently: you can try to remove `toolchain` in this go.mod and run `go version` (or any other command). You'll see this error message:

```sh
go: downloading go1.22 (darwin/amd64)
go: download go1.22 for darwin/amd64: toolchain not available
```

test/schema/default_config_test.go
dirkkul: Could you use the latest version from weaviate:stable/v1.26? We just renamed it there

go.mod
bevzzz: I changed it to `go 1.22.0` and removed the `toolchain` directive, because that was the simplest solution I could think of  -- let me know if that works for you

dirkkul (approved):

go.mod
dirkkul: yeah works :)

PR #246, opened 2024-10-29
Add support for multiple vectors per target

bevzzz:
This PR extends #239 by making sure `targetVectors: []` repeats the name of the target for each vector associated with it.


weaviate/graphql/nearvectorbuilder.go
```diff
@@ -106,9 +105,39 @@ func (b *NearVectorArgumentBuilder) build() string {
...
+//
+// Note, too, that in case the user fails to pass a value in TargetVectors,
+// it will not be added to the query.
+func (b NearVectorArgumentBuilder) prepareTargetVectors(targets []string) (out []string) {
+	for _, target := range targets {
+		if vectors, ok := b.vectorsPerTarget[target]; ok {
+			for range vectors {
+				out = append(out, target)
+			}
+			continue
```
dirkkul, line 138: can we do a 
```
if ok{
   ...repeated adding..
}else{
   out = append(out, target)
}
```

instead of doing the continue? Was a bit confused by this at first :)

dirkkul (approved):

PR #250, opened 2024-12-31
Add support for ColBERT embeddings

antas-marcin:


test/auth/auth_test.go
```diff
@@ -29,6 +29,7 @@ const (
 )
 
 func TestAuth_clientCredential(t *testing.T) {
+	t.Skip("Skipping bc client secret keys are expired")
```
tsmith023, line 32: Is this still true?

weaviate/graphql/nearvectorbuilder_test.go
```diff
@@ -78,7 +79,7 @@ func TestNearMultiVectorBuilder_build(t *testing.T) {
 
 	t.Run("No combination with vector per target", func(t *testing.T) {
 		vector := NearVectorArgumentBuilder{}
-		str := vector.WithVectorPerTarget(map[string][]float32{"one": {1, 2, 3}, "two": {4, 5, 6}}).build()
+		str := vector.WithVectorPerTarget(map[string]models.Vector{"one": []float32{1, 2, 3}, "two": []float32{4, 5, 6}}).build()
```
tsmith023, line 82: Does this change constitute a breaking change? If so, is it one that we have to tolerate due to the tight server/client coupling?

tsmith023 (approved):

test/auth/auth_test.go
antas-marcin: yes, it's true for all clients

weaviate/graphql/nearvectorbuilder_test.go
antas-marcin: yes, it's a breaking change, but a really simple one to resolve

PR #252, opened 2025-01-17
Align New with NewClient

taigrr:
The `New()` function doesn't take the Auth options into account, which is confusing, since the two functions are otherwise identical.

This PR unifies the logic, but preserves the `New()` function's `panic()` behavior and return signature for backwards compatibility.

### DIFF
--- a/weaviate/weaviate_client.go
+++ b/weaviate/weaviate_client.go
@@ -180,47 +180,12 @@ func NewClient(config Config) (*Client, error) {
 }
 
 // New client from config
-// Every function represents one API group of weaviate and provides a set of functions and builders to interact with them.
-//
-// The client uses the original data models as provided by weaviate itself.
-// All these models are provided in the sub module "github.com/weaviate/weaviate/entities/models"
+// For backwards compatibility, this function will panic if the client cannot be created
 func New(config Config) *Client {
-	con := connection.NewConnection(config.Scheme, config.Host, config.ConnectionClient, config.getTimeout(), config.Headers)
-
-	// some endpoints now require a className namespace.
-	// to determine if this new convention is to be used,
-	// we must check the weaviate server version
-	getVersionFn := func() string {
-		meta, err := misc.New(con, nil).MetaGetter().Do(context.Background())
-		if err == nil {
-			return meta.Version
-		}
-		return ""
-	}
-
-	dbVersionProvider := db.NewVersionProvider(getVersionFn)
-	dbVersionSupport := db.NewDBVersionSupport(dbVersionProvider)
-	gRPCVersionSupport := db.NewGRPCVersionSupport(dbVersionProvider)
-
-	grpcClient, err := createGrpcClient(config, gRPCVersionSupport)
+	client, err := NewClient(config)
 	if err != nil {
 		panic(err)
 	}
-
-	client := &Client{
-		connection:      con,
-		grpcClient:      grpcClient,
-		misc:            misc.New(con, dbVersionProvider),
-		schema:          schema.New(con),
-		c11y:            contextionary.New(con),
-		classifications: classifications.New(con),
-		graphQL:         graphql.New(con),
-		data:            data.New(con, dbVersionSupport),
-		batch:           batch.New(con, grpcClient, dbVersionSupport),
-		backup:          backup.New(con),
-		cluster:         cluster.New(con),
-	}
-
 	return client
 }
 


taigrr:
I've contributed to verba in the past but if there's anything i need to sign (CLA, etc) please let me know

taigrr:
i agree

bevzzz:
Yes, I agree, that is definitely a confusing distinction between the two, and unnecessary by the looks of it. Good catch 👍 

I am wondering, however, if this wouldn't be a breaking change, in a way. I.e., someone's `New()` might now panic because of an authorization issue -- something it didn't do before. WDYT @dirkkul?

`NewClient` also waits for the server to startup (`New` doesn't), but that shouldn't be an issue.

dirkkul:
> I am wondering, however, if this wouldn't be a breaking change, in a way. I.e., someone's New() might now panic because of an authorization issue -- something it didn't do before. WDYT @dirkkul?

Could you check what was happening before in this case? I can't remember.

dirkkul:
I think we should add a deprecation notice to New(), I thought we did that a while back, but apparently not

bevzzz:
> Could you check what was happening before in this case?

As @taigrr mentioned, `New()` just ignored authorization altogether, so I imagine any unauthorized clients will just be getting errors later when trying to query data.

> I think we should add a deprecation notice to New()

Agreed, document and deprecate would be the best way to resolve this. @taigrr would you like to update the PR, or should I just do that in a new one?

taigrr:
Unfortunately many libraries I've seen are using the New() call and won't work anymore because they throw unauthorized error panics. I've mentioned one here, langchaingo.

I presume they used to work, and the authorization flow here on the weaviate side was changed, and the breaking change is not my PR but rather some change that happened a ways back. 

For the sake of those other libraries I'd ask you to consider the change I presented once more, since in their current state they're completely broken, and a go get -u could fix them back to working status.

However, a deprecation notice also makes sense for new users going forward, since calling panic on connect is not great, returning an error is clearly better.

As an aside, the behavior is the same for unauthorized clients. (I apologize for formatting, responding from GitHub mobile)
This code panics:

```
        client = weaviate.New(weaviate.Config{
                Host:       weaviateURL,
                Scheme:     "https",
                AuthConfig: auth.ApiKey{Value: os.Getenv(vars.WeaviateAPIKey)},
                Headers: map[string]string{
                        "X-Azure-Api-Key": os.Getenv(vars.AZOAIVectorToken),
                },
        })
  
```
I provided an AuthConfig, and it was ignored, and the call to New() threw a panic as a result.



If you still disagree, that's fine, I understand, just close my PR and make the new one. I'd just have to maintain my fork for now.

bevzzz:
Hey @taigrr, thanks for providing more details here. I'm going to take another look at it tomorrow to see what the best solution would be.

Do I understand you correctly that libs like `langchaingo` are creating a client using the `New()` method, which _panics immediately_ because it skips authorization and then cannot create a connection to Weaviate server?

taigrr:
To the User of langchaingo uses the weaviate vector provider, and passes it to an LLM tool array. On first execution calling the tool, the plugin attempts to search, weaviate responds with a 401 unauthorized error when the toolcalls run. Following their example code, this triggers a runtime panic, the usage of the weavaite client is opaque post-initialization.

Additionally, if you try to provision schema in the init() call, which most will do, init() can't return an error, and most people panic here.

New() itself of course doesn't panic, it just creates the client offline without actually connecting

bevzzz:
hi again @taigrr!
After some deliberation, it seems that the best way to move forward would be to preserve the old code in `New()` as-is, but have it call `NewClient()` in the very beginning and return the created Client if there was no error.

This way a Client with a valid Auth config will be correctly authenticated whether it was created with `New` or `NewClient`.
In case the Auth config is invalid and `NewClient` returns an error we will retain the current behavior of getting a 401 when sending a config.

```go
func New(c Config) *Client {
    if client, err := NewClient(c); err == nil {
        return client
    }
    // ... current code of New()
}
```

taigrr:
That sounds great. I've updated the PR and rebased to match

weaviate/weaviate_client.go
```diff
@@ -180,11 +180,13 @@ func NewClient(config Config) (*Client, error) {
 }
 
 // New client from config
-// Every function represents one API group of weaviate and provides a set of functions and builders to interact with them.
-//
-// The client uses the original data models as provided by weaviate itself.
-// All these models are provided in the sub module "github.com/weaviate/weaviate/entities/models"
+// For backwards compatibility, this function will panic if the client cannot be created
```
bevzzz, line 183:
```suggestion
```
I think adding this line is unnecessary, because we won't be updating any code related to the panic

bevzzz:
Thank you for your contribution @taigrr 🙌

Issue #248, opened 2024-12-31
Add support for RBAC


dirkkul:
see python client for implementation

PR #255, opened 2025-01-28
RBAC Support

bevzzz:
Related #248


weaviate/rbac/role_all_getter.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/weaviate/weaviate-go-client/v4/weaviate/connection"
+	"github.com/weaviate/weaviate-go-client/v4/weaviate/except"
+	"github.com/weaviate/weaviate/entities/models"
+)
+
+type RoleAllGetter struct {
+	connection *connection.Connection
+}
+
+func (rag *RoleAllGetter) Do(ctx context.Context) ([]*models.Role, error) {
```
dirkkul, line 16: not the biggest fan of using models.Role as return because of the *string in there

test/testsuit/generics.go
```diff
@@ -197,16 +205,27 @@ func createWeaviateTestSchemaFoodWithReferenceProperty(t *testing.T, client *wea
...
+	}
 }
 
 // CreateTestClient running on local host 8080
 func CreateTestClient(enableGRPC bool) *weaviate.Client {
 	port, grpcPort, authEnabled := GetPortAndAuthPw()
 
-	openAIApiKey := os.Getenv("OPENAI_APIKEY")
 	headers := map[string]string{}
 	if openAIApiKey != "" {
```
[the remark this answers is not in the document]
bevzzz: Do you mean the headers do not need to be set here? If so, then why?

I deleted that line because I made it so that `openAIApiKey` would be read once at the top on the file

test/testsuit/generics.go
```diff
@@ -229,14 +248,39 @@ func CreateTestClient(enableGRPC bool) *weaviate.Client {
 		cfg.AuthConfig = auth.ApiKey{Value: "my-secret-key"}
 		client, err = weaviate.NewClient(cfg)
 		if err != nil {
-			log.Printf("Error occurred during startup %v", err)
```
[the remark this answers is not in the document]
bevzzz: Must've accidentally deleted that while cleaning up some of my debug prints, good catch!

weaviate/rbac/role_revoker.go
```diff
@@ -0,0 +1,45 @@
...
+	roles []string
+}
+
+func (rr *RoleRevoker) WithUser(user string) *RoleRevoker {
+	rr.user = user
+	return rr
+}
+
+func (rr *RoleRevoker) WithRoles(roles ...string) *RoleRevoker {
+	rr.roles = append([]string(nil), roles...)
```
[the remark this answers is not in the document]
bevzzz: Yeah, probably not :)

weaviate/rbac/user_roles_getter.go
```diff
@@ -0,0 +1,43 @@
...
+
+	user string
+}
+
+func (urg *UserRolesGetter) WithUser(user string) *UserRolesGetter {
+	urg.user = user
+	return urg
+}
+
+func (urg *UserRolesGetter) Do(ctx context.Context) ([]*models.Role, error) {
```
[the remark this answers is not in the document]
bevzzz: Updated 👌

test/rbac/rbac_test.go
```diff
@@ -0,0 +1,185 @@
...
+			"only %q should be assigned to %q", adminUser, adminRole)
+	})
+
+	t.Run("create role", func(t *testing.T) {
+		roleName := "TestRole"
+
+		mustCreateRole(t, roleName, &models.Permission{
+			Action:  &manageBackups,
+			Backups: &models.PermissionBackups{Collection: &pizza},
+		})
```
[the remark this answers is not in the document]
bevzzz: > We have that for example for the hybrid arguments

Shoot, I just noticed the second part of that comment. Sorry! Could you please let me know if the syntax I used is a good alternative?

weaviate/rbac/permission.go
```diff
@@ -0,0 +1,284 @@
...
+	}, actions...)
+}
+
+func ClusterPermissions(actions ...string) PermissionGroup {
+	return newPermissionGroup(func(action string) RoleOption {
+		return ClusterPermission{Action: action}
+	}, actions...)
+}
+
+func CollectionsPermissions(collection string, actions ...string) PermissionGroup {
```
dirkkul, line 47: In general I like this approach, but I don't think this will work for go. The problem being that we will have a tenant filter soonish and then need to change this to
`CollectionsPermissions(collection, tenant string, actions ...string) PermissionGroup` which means a breaking change for go :( Sadly I think we need to have a builder

(I don't understand why they do not have optional arguments, this makes no sense to me)

bevzzz: Noted! Will update today.

test/stop_containers.sh
```diff
@@ -4,11 +4,8 @@ set -eou pipefail
...
 
-docker compose -f test/docker-compose.yml down --remove-orphans
-docker compose -f test/docker-compose-azure.yml down --remove-orphans
-docker compose -f test/docker-compose-okta-cc.yml down --remove-orphans
-docker compose -f test/docker-compose-okta-users.yml down --remove-orphans
-docker compose -f test/docker-compose-wcs.yml down --remove-orphans
-docker compose -f test/docker-compose-cluster.yml down --remove-orphans
+# Removes containers in the default docker-compose.yml
+# and _any other containers_ not defined in that file.
+docker compose down --remove-orphans
```
antas-marcin, line 9: question: this will tear down all of the docker compose setups?

weaviate/rbac/role_getter.go
```diff
@@ -0,0 +1,156 @@
...
+					Collection: resources[0],
+					Verbosity:  resources[1],
+				}
+			}, *perm.Action, *perm.Nodes.Collection, *perm.Nodes.Verbosity)
+			// Scope comes back as `nil` if not set.
+		case perm.Roles != nil:
+			// scope := ""
+			// if (perm.Roles.Scope != nil) {
+			// 	scope = *perm.Roles.Scope
+			// }
```
antas-marcin, line 85: shouldn't this part be removed?

weaviate/users/my_user_getter.go
```diff
@@ -0,0 +1,28 @@
...
+func (mug *MyUserGetter) Do(ctx context.Context) (*models.UserInfo, error) {
+	path := "/users/own-info"
+	res, err := mug.connection.RunREST(ctx, path, http.MethodGet, nil)
+	if err != nil {
+		return nil, except.NewDerivedWeaviateClientError(err)
+	}
+	if res.StatusCode == http.StatusOK {
+		var user *models.UserInfo
+		decodeErr := res.DecodeBodyIntoTarget(&user)
+		return user, decodeErr
```
antas-marcin, line 25: question: why do we need to pass a pointer to a nil variable? can we instead pass just `nil`?

test/stop_containers.sh
bevzzz: Yes, I tried locally with starting several "groups" of containers -- running this command once removes all of them.

Because it's `remove all containers in docker-compose.yml` + `remove any container that's not in the docker-compose.yml` -- effectively **all** containers

weaviate/users/my_user_getter.go
bevzzz: We can either have `var user *models.UserInfo` or `var user models.UserInfo`, but we'll need to pass `&user` to `json.Unmarshal` either way.

Here's a [Go Playground example](https://go.dev/play/p/Fxc0XUdQL3k) and the [SO post](https://stackoverflow.com/questions/35604356/json-unmarshal-accepts-a-pointer-to-a-pointer) I learned about this from

bevzzz: In general, do you think clients should returns structs or pointers to structs?
I know returning pointers may save some memory, but in this case it doesn't seem like we're going to win much.

Maybe we should return `models.UserInfo` here instead of `*models.UserInfo`.

antas-marcin: got it, didn't know that we inside an Unmarshaler

weaviate/rbac/role_getter.go
```diff
@@ -0,0 +1,151 @@
...
+	backups := make(mergedPermissions)
+	collections := make(mergedPermissions)
+	data := make(mergedPermissions)
+	nodes := make(mergedPermissions)
+	roles := make(mergedPermissions)
+	clusters := make(mergedPermissions)
+	tenants := make(mergedPermissions)
+	users := make(mergedPermissions)
+
+	for _, perm := range r.Permissions {
```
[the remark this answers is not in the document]
bevzzz: Thanks for reminding me about that -- found a bug while adding unit tests for that :)

weaviate/users/role_assigner.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+type RoleAssigner struct {
+	connection *connection.Connection
+
+	user  string
+	roles []string
+}
+
+func (ra *RoleAssigner) WithUser(user string) *RoleAssigner {
```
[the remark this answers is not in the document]
bevzzz: Updated 👌

weaviate/users/my_user_getter.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/weaviate/weaviate-go-client/v4/weaviate/connection"
+	"github.com/weaviate/weaviate-go-client/v4/weaviate/except"
+	"github.com/weaviate/weaviate/entities/models"
+)
+
+type MyUserGetter struct {
+	connection *connection.Connection
+}
+
+func (mug *MyUserGetter) Do(ctx context.Context) (*models.UserInfo, error) {
```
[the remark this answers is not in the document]
bevzzz: @dirkkul could you please take a look at [28b1aab](https://github.com/weaviate/weaviate-go-client/pull/255/commits/28b1aabd4a75e0ab3b50033e1dd999284419fd3f) and let me know if that's what you had in mind?

weaviate/users/my_user_getter.go
```diff
@@ -33,3 +33,25 @@ func (mug *MyUserGetter) Do(ctx context.Context) (*models.UserInfo, error) {
 	}
 	return nil, except.NewUnexpectedStatusCodeErrorFromRESTResponse(res)
 }
+
+func (info *UserInfo) UnmarshalJSON(data []byte) error {
```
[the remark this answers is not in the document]
bevzzz: Oh yes I did, nice catch, thank you!

dirkkul (approved):

PR #256, opened 2025-02-25
Add necessary headers for weaviate embedding service if present

tsmith023:


.gitignore
```diff
@@ -9,4 +9,5 @@ tmp
 vendor/
 .vscode/
 .idea/
-.env
\ No newline at end of file
+.env
+test/scratch_test.go
```
bevzzz, line 13: suggestion: let's check that test in, but `t.Skip("only run locally")` it right away.
Maybe read the hosts + API keys from environment so as not to expose them.

Then the test could be re-run another time without rewriting it.

bevzzz (approved):

PR #259, opened 2025-04-02
Support Dynamic User Management in Weaviate v1.30

mpartipilo:
Update Weaviate Go client with Dynamic User Management features introduced in Weaviate v1.30.

* Dedicated scopes introduced for DB and OIDC users, via `client.Users().DB()` and `client.Users().OIDC()`
* New information return for Users:
  * Active Status
  * User Types (DB, DB Env, OIDC)
  * Creation Time
* DB User Creation / Deletion
* DB User Deactivation / Activation
* DB User API Key Rotation
* Listing of users assigned to a role


weaviate/users/user_db_lister.go
```diff
@@ -0,0 +1,42 @@
...
+	"net/http"
+
+	"github.com/weaviate/weaviate-go-client/v5/weaviate/connection"
+	"github.com/weaviate/weaviate-go-client/v5/weaviate/except"
+	"github.com/weaviate/weaviate/client/users"
+)
+
+type UserInfoList []UserInfo
+
+type UserDbLister struct {
```
bevzzz, line 14:
```suggestion
type UserDBLister struct {
```

suggestion: [Initialisms](https://google.github.io/styleguide/go/decisions.html#initialisms) should have consistent case

Not to spam the PR with identical comments, also applies to UserDbActivator, UserDbDeactivator, UserDbGetter, etc.

weaviate/users/users.go
```diff
@@ -1,6 +1,16 @@
...
-import "github.com/weaviate/weaviate-go-client/v5/weaviate/connection"
+import (
+	"github.com/weaviate/weaviate-go-client/v5/weaviate/connection"
+	"github.com/weaviate/weaviate/entities/models"
+)
+
+type UserType string
+
+const (
+	UserTypeDb   UserType = UserType(models.UserTypeDb)
```
bevzzz, line 11:
```suggestion
	UserTypeDB   UserType = UserType(models.UserTypeDb)
```
Same as the note on initialisms above ☝️ 

I see that `models` doesn't adhere to that, but we can maintain that style in the client regardless, just like we do below with `UserTypeOIDC`

weaviate/users/user_roles_getter.go
```diff
@@ -13,8 +13,9 @@ import (
 type UserRolesGetter struct {
 	connection *connection.Connection
 
-	userID   string
-	userType string
+	userID           string
+	userType         string
+	IncludeFullRoles *bool
```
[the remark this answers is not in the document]
mpartipilo: Good catch. It's not supposed to be exported.

test/users/users_test.go
```diff
@@ -48,7 +48,7 @@ func TestUsers_integration(t *testing.T) {
 	}
 
 	t.Run("get user roles", func(t *testing.T) {
-		adminRoles, err := usersClient.UserRolesGetter().WithUserID(adminUser).Do(ctx)
+		adminRoles, err := usersClient.DB().RolesGetter().WithUserID(adminUser).WithIncludeFullRoles(true).Do(ctx)
```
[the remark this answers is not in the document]
mpartipilo: Makes sense

weaviate/users/user_db_creator.go
```diff
@@ -0,0 +1,39 @@
...
+	userID string
+}
+
+func (r *UserDBCreator) WithUserID(id string) *UserDBCreator {
+	r.userID = id
+	return r
+}
+
+func (r *UserDBCreator) Do(ctx context.Context) error {
+	payload := users.NewCreateUserParams().WithUserID(r.userID)
```
dirkkul, line 25: should we add a runtime check that user_id is given? (same for the other endpoints that have required parameters)

weaviate/users/user_db_deleter.go
```diff
@@ -0,0 +1,39 @@
...
+
+	userID string
+}
+
+func (r *UserDBDeleter) WithUserID(id string) *UserDBDeleter {
+	r.userID = id
+	return r
+}
+
+func (r *UserDBDeleter) Do(ctx context.Context) error {
```
dirkkul, line 24: in python we return a bool to indicate if the operation deleted the user or not, so we could do

```
func (r *UserDBDeleter) Do(ctx context.Context) (bool,error) {
	res, err := r.connection.RunREST(ctx, r.path(), http.MethodDelete, payload)
	if err != nil {
		return except.NewDerivedWeaviateClientError(err)
	}
	if res.StatusCode == http.StatusNoContent {
		return true, nil
	} else if if res.StatusCode == http.StatusConflict{. // please double check this one
               return false, nil
        }
	return except.NewUnexpectedStatusCodeErrorFromRESTResponse(res)
}
```

weaviate/users/user_roles_getter.go
```diff
@@ -13,15 +13,31 @@ import (
...
+func (urg *UserRolesGetter) WithIncludeFullRoles(include bool) *UserRolesGetter {
+	urg.includeFullRoles = include
+	return urg
+}
+
 func (urg *UserRolesGetter) Do(ctx context.Context) ([]*rbac.Role, error) {
+	// Assume DB user if no user type is specified
+	if urg.userType == "" {
+		urg = urg.WithUserType(UserTypeDB)
+	}
```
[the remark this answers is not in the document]
dirkkul: Just to be clear:
users.get_roles() => GET to "/authz/users/%s/roles"

users.db.get_roles() => GET "/authz/users/%s/roles/DB"
users.oidc.get_roles() => GET "/authz/users/%s/roles/OIDC"

weaviate/users/role_revoker.go
```diff
@@ -27,10 +29,19 @@ func (rr *RoleRevoker) WithRoles(roles ...string) *RoleRevoker {
...
+
 func (rr *RoleRevoker) Do(ctx context.Context) error {
-	res, err := rr.connection.RunREST(ctx, rr.path(), http.MethodPost, authz.RevokeRoleFromUserBody{
+	payload := authz.RevokeRoleFromUserBody{
 		Roles: rr.roles,
-	})
+	}
+	if rr.userType != "" {
+		payload.UserType = models.UserType(rr.userType)
+	}
```
dirkkul, line 43: same as with the roles getter

weaviate/users/role_assigner.go
```diff
@@ -27,10 +29,19 @@ func (ra *RoleAssigner) WithRoles(roles ...string) *RoleAssigner {
...
+
 func (ra *RoleAssigner) Do(ctx context.Context) error {
-	res, err := ra.connection.RunREST(ctx, ra.path(), http.MethodPost, authz.AssignRoleToUserBody{
+	payload := authz.AssignRoleToUserBody{
 		Roles: ra.roles,
-	})
+	}
+	if ra.userType != "" {
+		payload.UserType = models.UserType(ra.userType)
+	}
```
dirkkul, line 43: same as with the roles getter

go.mod
```diff
@@ -9,10 +9,10 @@ require (
 	github.com/stretchr/testify v1.10.0
 	github.com/testcontainers/testcontainers-go v0.35.0
 	github.com/testcontainers/testcontainers-go/modules/weaviate v0.35.0
-	github.com/weaviate/weaviate v1.29.0
+	github.com/weaviate/weaviate v1.30.0-rc.0
```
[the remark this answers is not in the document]
mpartipilo: Got it. Already fixed the breaking change issues.

bevzzz:
Should we add some minimal tests for db-only actions: create/delete users, rotate key, de-/activate?

weaviate/users/user_db_deleter.go
mpartipilo: Funny, I had it like this before and eventually decided to make it more go-like by assuming success on lack of error. I'll add the boolean back.

weaviate/users/user_db_creator.go
bevzzz: Seems we don't do it for (most?) of the other endpoints throughout the codebase, e.g. DELETE /schema/{class}

https://github.com/weaviate/weaviate-go-client/blob/9864f0c0d99ad07a9880c364cd76d12e61354c93/weaviate/schema/classDeleter.go#L25-L26

so there isn't a well-established pattern yet.
Is it worth introducing one for /users/... endpoints?

mpartipilo: I would have no issue adding parameter validation to all endpoints in a separate PR.

weaviate/users/user_db_deleter.go
mpartipilo: The other status returned is 422, UnprocessableEntity. I don't see a 409 returned by the rest handler for Delete.

dirkkul: ahh it is 404

dirkkul:
```
      "delete": {
        "summary": "Delete User",
        "operationId": "deleteUser",
        "x-serviceIds": [
          "weaviate.users.db.delete"
        ],
        "tags": [
          "users"
        ],
        "parameters": [
          {
            "description": "user name",
            "in": "path",
            "name": "user_id",
            "required": true,
            "type": "string"
          }
        ],
        "responses": {
          "204": {
            "description": "Successfully deleted."
          },
          "400": {
            "description": "Malformed request.",
            "schema": {
              "$ref": "#/definitions/ErrorResponse"
            }
          },
          "401": {
            "description": "Unauthorized or invalid credentials."
          },
          "403": {
            "description": "Forbidden",
            "schema": {
              "$ref": "#/definitions/ErrorResponse"
            }
          },
          "404": {
            "description": "user not found"
          },
          "422": {
            "description": "Request body is well-formed (i.e., syntactically correct), but semantically erroneous.",
            "schema": {
              "$ref": "#/definitions/ErrorResponse"
            }
          },
          "500": {
            "description": "An error has occurred while trying to fulfill the request. Most likely the ErrorResponse will contain more information about the error.",
            "schema": {
              "$ref": "#/definitions/ErrorResponse"
            }
          }
        }
      }
    },
```

mpartipilo: Thanks!

dirkkul: > Funny, I had it like this before and eventually decided to make it more go-like by assuming success on lack of error. I'll add the boolean back.

I mean both cases are "success" in the sense that the user is gone. Sometimes you care about if the user was there before and sometimes you don't :) I think I like the idea of deletes being idempotent, but have the bool in case you want to be sure

weaviate/users/user_db_deactivator.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/weaviate/weaviate/client/users"
+)
+
+type UserDBDeactivator struct {
+	connection *connection.Connection
+
+	userID string
+}
+
+func (r *UserDBDeactivator) WithUserID(id string) *UserDBDeactivator {
```
dirkkul, line 19: There is a "revoke_key" option that is missing (I also forgot in the python client)

bevzzz (dismissed):

weaviate/users/user_operations_db.go
```diff
@@ -0,0 +1,43 @@
...
+type UserOperationsDB struct {
+	connection *connection.Connection
+}
+
+func (r *UserOperationsDB) RolesGetter() *UserRolesGetter {
+	return (&UserRolesGetter{connection: r.connection}).WithUserType(UserTypeDB)
+}
+
+func (r *UserOperationsDB) RolesAssigner() *RoleAssigner {
+	return (&RoleAssigner{connection: r.connection}).WithUserType(UserTypeDB)
```
bevzzz, line 14: A user can still override the user type via WithUserType later -- is that an issue?
This lets us keep code duplication at a minimum and you can never make stuff entirely fool-proof; could we make `withUserType()` an unexported method?

bevzzz (changes requested):
I think there needs to be one more change in the `rbac.AssignedUsersGetter`.
Namely, that it should hit a different endpoint and return `{user_id: string, user_type: string}[]` instead of `string[]` as it does now.

See the related change in Python client:

https://github.com/weaviate/weaviate-python-client/blob/366d4f82ac34a0be4bbef32b826e7cbeca163d28/weaviate/rbac/roles.py#L165

Ideally, we should deprecate the current AssignedUsersGetter and add sth like UserAssignments getter to avoid breaking the existing API.

___

ps: that's probably what Dirk was talking about [here](https://github.com/weaviate/weaviate-go-client/pull/259#pullrequestreview-2738929075).

bevzzz:
> I think there needs to be one more change in the `rbac.AssignedUsersGetter`. 

Ideally, we should deprecate the current `AssignedUsersGetter` and add sth like UserAssignments getter to avoid breaking the existing API.

mpartipilo:
> I think there needs to be one more change in the `rbac.AssignedUsersGetter`. Namely, that it should hit a different endpoint and return `{user_id: string, user_type: string}[]` instead of `string[]` as it does now.
> 
> See the related change in Python client:
> 
> [weaviate/weaviate-python-client@`366d4f8`/weaviate/rbac/roles.py#L165](https://github.com/weaviate/weaviate-python-client/blob/366d4f82ac34a0be4bbef32b826e7cbeca163d28/weaviate/rbac/roles.py#L165)
> 
> Ideally, we should deprecate the current AssignedUsersGetter and add sth like UserAssignments getter to avoid breaking the existing API.
> 
> ps: that's probably what Dirk was talking about [here](https://github.com/weaviate/weaviate-go-client/pull/259#pullrequestreview-2738929075).

Yes, I didn't spot that change. Good catch, thanks!

test/users/users_test.go
```diff
@@ -73,13 +88,39 @@ func TestUsers_integration(t *testing.T) {
...
+			func(e rbac.UserAssignment) bool { return e.UserID == adminUser }), "should have %q role", roleName)
+
+		// Act: revoke
+		err = usersClient.DB().RolesRevoker().WithUserID(adminUser).WithRoles(roleName).Do(ctx)
+		require.NoErrorf(t, err, "revoke %q role", roleName)
+
+		assignedUsers, _ = rolesClient.AssignedUsersGetter().WithRole(roleName).Do(ctx)
+		require.Falsef(t, slices.ContainsFunc(assignedUsers,
+			func(e rbac.UserAssignment) bool { return e.UserID == adminUser }), "should not have %q role", roleName)
 	})
```
dirkkul, line 125: could you add simple tests for all the other functions?

dirkkul (approved):

PR #263, opened 2025-06-02
Support new GraphQL search operators (minimumOrTokensMatch) for bm25/hybrid query

bevzzz:
This PR adds `BM25SearchOperatorBuilder` to add the new search operator to the GraphQL query and the corresponding builder methods to BM25/Hybrid query builders.

Tested only using unit tests, because correct GraphQL query is the key here.


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/weaviate/weaviate-go-client/v5
 
-go 1.22.7
+go 1.24
+
+toolchain go1.24.1
```
dirkkul, line 5: do we need to change this? I think that would force our users to update

bevzzz: I just thought the same and reverted these changes 👍

dirkkul (approved):

PR #268, opened 2025-06-30
feat(alias): Add support for Collection Alias in Go client

kavirajk:
This PR adds client support for handling recent alias support.

Refer [this PR ](https://github.com/weaviate/weaviate/pull/8483)on the server side to get the idea of all the endpoints.


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/weaviate/weaviate-go-client/v5
 
-go 1.23.0
+go 1.24
+
+toolchain go1.24.4
```
antas-marcin, line 5: can we remove the toolchain directive?

go.mod
```diff
@@ -9,12 +11,12 @@ require (
 	github.com/stretchr/testify v1.10.0
 	github.com/testcontainers/testcontainers-go v0.35.0
 	github.com/testcontainers/testcontainers-go/modules/weaviate v0.35.0
-	github.com/weaviate/weaviate v1.30.0
+	github.com/weaviate/weaviate v1.31.4-0.20250627212219-1140d3c8a50e
```
antas-marcin, line 14: you can update Weaviate dependency to `v1.32.0-rc.0`

weaviate/alias/alias.go
```diff
@@ -0,0 +1,48 @@
...
+	connection *connection.Connection
+}
+
+// New alias api group from connection
+func New(con *connection.Connection) *API {
+	return &API{connection: con}
+}
+
+// Getter builder to get a weaviate aliases
+func (schema *API) AliasLister() *AliasList {
```
antas-marcin, line 18: let's rename this to `AliasGetter`

weaviate/alias/getter.go
```diff
@@ -0,0 +1,49 @@
...
+	s.className = strings.TrimSpace(className)
+	return s
+}
+
+// Do get the list of alias
+func (s *Getter) Do(ctx context.Context) ([]Alias, error) {
+	return listAlias(ctx, s.connection, s.className)
+}
+
+func listAlias(ctx context.Context, conn *connection.Connection, className string) ([]Alias, error) {
```
antas-marcin, line 30: can we make this method a part of `Getter` struct?

kavirajk: well. it doesn't have to be. Just simple func needing two arguments :). It doesn't need anything else from the struct. Even by making it part of struct, it makes sense to pass these args explicitly.

PR #277, opened 2025-08-13
feat(rbac): add Replicate and Alias permissions

bevzzz:
This PR adds `AliasPermission` and `ReplicatePermissions` to manage corresponding permissions in the RBAC.


dirkkul (commented):
Could you extend the unit test in https://github.com/weaviate/weaviate-go-client/blob/main/weaviate/rbac/permission_test.go?

PR #279, opened 2025-08-18
3/4 feature: introduces `Not` operator

aliszka:


antas-marcin (approved):

weaviate/filters/where.go
```diff
@@ -216,7 +216,9 @@ func (b *WhereBuilder) ToGRPC() *pb.Filters {
 		filters.Operator = pb.Filters_OPERATOR_AND
 		filters.Filters = operandsToGRPC(b.operands)
 
-	// TODO aliszka add Not
+	case Not:
+		filters.Operator = pb.Filters_OPERATOR_NOT
+		filters.Filters = operandsToGRPC(b.operands)
```
bevzzz, line 221: I remember you mentioning that `Not` expects only 1 child operand like Not(Like("%this%")) -- do we want to ensure that somehow here? E.g. `operandsToGRPC(b.operands[0])` (+ bounds check)?

aliszka: so far validation was done on db side, clients were pushing to the db everything user has added to the request.

what would be you suggestion here? let db do the validation? let client do the validation and return error on len != 1? or silently ignore additional operands?

bevzzz: I agree that all validation that can be done by DB should be delegated to DB 👍 

I was thinking along the lines of only including the 1st operand (see my comment above), to guardrail incorrect usage. But that has the effect of silencing that incorrect usage and may not be expected by the user.

Now that I've thought about it again, your original version is good, as the users will quickly get a validation error from the DB should they misuse the NOT operator.

Resolved!

PR #282, opened 2025-08-22
Grpc filters

aliszka:


antas-marcin (approved):

weaviate/filters/where.go
```diff
@@ -210,6 +204,112 @@ func (b *WhereBuilder) string() string {
...
+		filters.Filters = operandsToGRPC(b.operands)
+
+	case And:
+		filters.Operator = pb.Filters_OPERATOR_AND
+		filters.Filters = operandsToGRPC(b.operands)
+
+	// TODO aliszka add Not
+
+	default:
+		//nolint:staticcheck // SA1019
```
bevzzz, line 222: What kind of lint error does this silence?

aliszka: `filters.On` is deprecated, but since client's api allows only to provide path this way, deprecated method have to be used for now.
will add another comment with explanation

dirkkul: Hey - `FilterTarget target = 20` is the correct way to provide the target. I am about to remove `On`- could you please switch it over?

PR #369, opened 2026-04-09
v6: Count objects in collection

bevzzz:
This PR adds `handle.Count(ctx)` helper function, similar to Python's `len(collection)` and Java's `handle.size()`, as a shorthand for the "count(*)"-aggregation.

### DIFF
--- a/internal/api/aggregate.go
+++ b/internal/api/aggregate.go
@@ -1,11 +1,7 @@
 package api
 
 import (
-	"cmp"
 	"fmt"
-	"iter"
-	"maps"
-	"slices"
 	"time"
 
 	proto "github.com/weaviate/weaviate-go-client/v6/internal/api/internal/gen/proto/v1"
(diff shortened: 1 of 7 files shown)


internal/api/aggregate.go
```diff
@@ -426,3 +439,54 @@ func nilZero[T comparable](v T) *T {
...
+// similar to [maps.All], but with pairs sorted by key.
+func sortedMap[Map ~map[K]V, K cmp.Ordered, V any](m Map) iter.Seq2[K, V] {
+	return func(yield func(K, V) bool) {
+		for _, k := range slices.Sorted(maps.Keys(m)) {
+			if !yield(k, m[k]) {
+				return
+			}
+		}
+	}
+}
```
tsmith023: Is this unused?

tsmith023 (approved):

internal/api/aggregate.go
bevzzz: as a fork in a pizza joint!

bevzzz: that was my attempt at a cheeky simile

PR #375, opened 2026-04-14
Add MCP permission

g-despot:
Add MCP RBAC permissions: `create_mcp`, `read_mcp` and `update_mcp`.


go.mod
```diff
@@ -1,34 +1,34 @@
 module github.com/weaviate/weaviate-go-client/v5
 
-go 1.25
+go 1.26
```
bevzzz, line 4: issue: we should not upgrade the project to 1.26 just yet, as we may have users who are still on 1.25

.github/workflows/workflow-tests.yaml
```diff
@@ -24,7 +24,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v6
         with:
-          go-version: "1.25"
+          go-version: "1.26"
           cache: true
```
bevzzz, line 28: issue: same as with `go.mod`, we want to stick to `1.25`

bevzzz (changes requested):

bevzzz (approved):

PR #380, opened 2026-04-20
v6: Journey tests in the CI

bevzzz:
This PR adds an `./example` directory, which will house the client's journey tests. Each journey test is structured as a standalone Go module, such that it can be ran with `go run ./example/<name>`.

The directory is called "examples" rather than "journey" or "journey tests" to highlight that it's something we'd like the users to look around and learn from -- it's a part of our documentation that also run in CI.

Apropos of CI, there is now an `examples.yaml` workflow, which will run all examples in parallel against a WCD instance. This should only run on push to `v6` (later to main) so as not to slow down the PR feedback loop. 

I added a simple example for now to set up the infrastructure around but I hope we'll expand that to 4-5 most frequent use-cases.

___

Misc: this also disables `contracts.yaml` workflow for PRs and only lets it run on push to v6. Otherwise a single update to the OpenAPI schema upstream can cause many PRs to go red for no good reason. Much ado about nothing.


example/go.mod
```diff
@@ -0,0 +1,43 @@
+module example/rbac
+
+go 1.26.1
+
+require github.com/weaviate/weaviate-go-client/v6 v6.0.0-example
```
tsmith023, line 5: I'm not seeing the value of having separate modules for each test as:

1. this will always need to be in sync between all modules, right? So is this not an opportunity for DRY by defining it once at the root?
2. `cd`-ing into the specific dir to call `go run .` is more work imo than staying in the root and calling `go run ./basic`

Is the idea that a prospective user would prefer `cd example/basic && go mod tidy && go run .` instead of `cd example && go mod tidy && go run ./basic`? Is the benefit of the former over the latter worth the cost of maintaining multiple `go.mod` files? wdyt?

bevzzz: I wholeheartedly agree with your suggestions, that hadn't crossed my mind before!
Lemme change that

tsmith023 (approved):

PR #383, opened 2026-04-23
v6: Filters

bevzzz:
This PR adds support for filter expressions in all supported queries. To avoid polluting the `query` namespaces all related types / functions are provided via a dedicated `query/filter` package. The expressions are composable and allow expressing complex filtering logic concisely.

~~The syntax for reference targets should be considered experimental: we might introduce a dedicated variadic `Reference` method and reserve `Property` to just single names and/or introduce method chaining similar to how it's done in the Python client.~~

The `query/filter` package trades clever for simple: rather than hide internal filter structure behind builders, we export a simple `Cond` struct with fields for Target, Operator, and Value. The `And`, `Or`, and `Not` types offer some syntactic sugar, but only insofar as the standard language syntax permits, e.g. And is a _slice_ of filter expressions.

For now, the goal is to keep both the API surface and the implementation small; we should revisit filters before GA, ideally with some real-world code snippets at hand.

### DIFF
--- a/internal/api/search.go
+++ b/internal/api/search.go
@@ -712,3 +712,21 @@ func (cl ConsistencyLevel) proto() *proto.ConsistencyLevel {
 		return nil
 	}
 }
+
+type FilterOperator proto.Filters_Operator
+
+const (
+	FilterOperatorNot              FilterOperator = FilterOperator(proto.Filters_OPERATOR_NOT)
+	FilterOperatorAnd              FilterOperator = FilterOperator(proto.Filters_OPERATOR_AND)
+	FilterOperatorOr               FilterOperator = FilterOperator(proto.Filters_OPERATOR_OR)
+	FilterOperatorEqual            FilterOperator = FilterOperator(proto.Filters_OPERATOR_EQUAL)
+	FilterOperatorLessThan         FilterOperator = FilterOperator(proto.Filters_OPERATOR_LESS_THAN)
+	FilterOperatorLessThanEqual    FilterOperator = FilterOperator(proto.Filters_OPERATOR_LESS_THAN_EQUAL)
+	FilterOperatorGreaterThan      FilterOperator = FilterOperator(proto.Filters_OPERATOR_GREATER_THAN)
+	FilterOperatorGreaterThanEqual FilterOperator = FilterOperator(proto.Filters_OPERATOR_GREATER_THAN_EQUAL)
+	FilterOperatorLike             FilterOperator = FilterOperator(proto.Filters_OPERATOR_LIKE)
+	FilterOperatorIsNull           FilterOperator = FilterOperator(proto.Filters_OPERATOR_IS_NULL)
+	FilterOperatorContainsAll      FilterOperator = FilterOperator(proto.Filters_OPERATOR_CONTAINS_ALL)
+	FilterOperatorContainsAny      FilterOperator = FilterOperator(proto.Filters_OPERATOR_CONTAINS_ANY)
+	FilterOperatorContainsNone     FilterOperator = FilterOperator(proto.Filters_OPERATOR_CONTAINS_NONE)
+)
--- a/internal/api/search.go
+++ b/internal/api/search.go
@@ -6,6 +6,7 @@ import (
 	"regexp"
 	"slices"
 	"time"
+	"unicode"
 
 	"github.com/google/uuid"
 	"github.com/weaviate/weaviate-go-client/v6/internal"
@@ -44,6 +45,12 @@ func (r *SearchRequest) Method() transport.MethodFunc[proto.SearchRequest, proto
 func (r *SearchRequest) Body() transport.MessageMarshaler[proto.SearchRequest] { return r }
 
 type (
+	Filter struct {
+		Operator FilterOperator
+		Exprs    []Filter // Sub-expressions of logical operators AND/OR/NOT.
+		Target   []string // Target of a comparison expression, i.e. property, reference.
+		Value    any      // Test value appropriate for the target data type.
+	}
 	ReturnMetadata struct {
 		CreatedAt    bool
 		LastUpdateAt bool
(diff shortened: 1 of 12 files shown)


query/filter/filter.go
```diff
@@ -0,0 +1,168 @@
...
+	//
+	// Example:
+	//
+	//	[]string{"album"} 									// Property "album" of the root Songs collection
+	//	[]string{"performedBy", "given_name"} 				// Property "given_name" of the referenced Artists collection (single-target reference)
+	//	[]string{"hasAwards", "GrammyAward", "year"} 		// Property "year" of the referenced Grammy collection (multi-target reference)
+	//	[]string{"performedBy", "bornIn", "population"}  	// performedBy -[Artists]-> bornIn -[Cities]-> population
+	//
+	// See [Len] and [ReferenceCount].
+	Target []string
```
tsmith023: Wdyt about also having functions like the following?
```go
func ByProperty(prop string) []string
...
ByProperty("album") => []string{"album"}
```
and
```go
func ByReferenceProperty(path ...string) []string
...
ByReferenceProperty("performedBy", "given_name") => []string{"performedBy", "given_name"}
...
ByReferenceProperty("performedBy", "bornIn", "population") => []string{"performedBy", "bornIn", "population"}
```
acting as helper equivalents to `func Len` and `func ReferenceCount`?

internal/api/search.go
```diff
@@ -263,6 +273,186 @@ func marshalReturnVectors(req *proto.MetadataRequest, vectors []string) {
...
+			pf.TestValue = &proto.Filters_ValueBoolean{ValueBoolean: v}
+		case []bool:
+			pf.TestValue = &proto.Filters_ValueBooleanArray{
+				ValueBooleanArray: &proto.BooleanArray{Values: v},
+			}
+
+		// Integer values must be cast to int64 before marshaling.
+		// structpb package supports uints, but those are probably
+		// quite rare in the use cases where Weaviate is used, so
+		// we ignore them.
```
tsmith023: Why not cast any `uintX` values to `int64`?

internal/api/search.go
```diff
@@ -263,6 +273,186 @@ func marshalReturnVectors(req *proto.MetadataRequest, vectors []string) {
...
+				},
+			}
+		} else {
+			return &proto.FilterTarget{
+				Target: &proto.FilterTarget_Property{
+					Property: property,
+				},
+			}
+		}
+	default:
```
tsmith023, line 474: This case will certainly need more validation in the future with acceptance tests in the case a user incorrectly provides their reference filter. We could rely on the server to send a nice error in such a case but that will need to be verified in the gRPC API layer to be 100% sure

internal/api/search.go
bevzzz: I was not sure if it'll be useful and wanted to avoid adding what I thought would be a rarely-used code path. Buuuut, I think you have a good point there and there's nothing in the API to say users shouldn't use uintX.

query/filter/filter.go
bevzzz: > Wdyt about also having functions like the following?
> ```go
> func ByProperty(prop string) []string
> func ByReferenceProperty(path ...string) []string
> ```

I understand that making the API more (I don't know what's the good word) prose-like can improve readability. However, exporting a function like the ones below doesn't sit right with me, as it is pure syntactic sugar with little to no functional utility. 

```go
func ByReferenceProperty(path ...string) []string { return path }
func ByProperty(path string) []string { return []string{path} }
```

I'd be rather surprised if I discovered something like that in some package.

> acting as helper equivalents to `func Len` and `func ReferenceCount`?

The difference with these is that Len and ReferenceCount hide an internal implementation detail, namely that Weaviate expects `len(property)` and not, say, `length(property)` and that `internal/api` package in the client expects `count(property)`. 

I find that a good extension to the `filter` package could be a `MultiReference` function which could use some special formatting, e.g. `"property:Collection"` that is understood internally but doesn't need to be exposed to the user.

internal/api/search.go
bevzzz: > This case will certainly need more validation in the future with acceptance tests 

Agreed, I'll make sure to use this in one of the [examples](#380), it'd be a useful reference (no pun intended)

> We could rely on the server to send a nice error

IMO it's the server that should be responsible for validating the reference, mainly because it has more information to return an instructive error message, e.g. 

```
no target collection specified for multi-reference property "hasAwards" (GrammyAwards, TonyAwards)
```

query/filter/filter.go
bevzzz: Thought about this some more and realized that the current design penalizes the most common by-property case by making the user pass `[]string{property}` instead of just `property`. We actually want to do the opposite, because references are less common.

I have two further options in mind. The first one presents different fields for property and reference targets, so that the common case is streamlined. We do a similar thing in query, where the more common ReturnProperties is `[]string` and ReturnNestedProperties is `[]ReturnProperty`. It leaves some room for error, but I think the naming mitigates it all.

```go
f := filter.Cond{Property: "name"}
f := filter.Cond{Reference: []string{"bestFriend", "name"}}

// Using with Len - OK
f := filter.Cond(Property: filter.Len("name")}
f := filter.Cond{Reference: []string{"bestFriend", filter.Len("name")}

// Using with ReferenceCount - idk, awkward
f := filter.Cond{Property: filter.ReferenceCount("hasFriends")}
```

Alternatively, Cond only has `Target string` field, and we provide a helper function for constructing reference targets.

```go
f := filter.Cond{Target: "name"}
f := filter.Cond{Target: filter.Reference("bestFriend", "name")}

// Using with ReferenceCount - OK
f := filter.Cond{Target: filter.ReferenceCount("hasFriends")}

// Using with Len + property - OK
f := filter.Cond{Target: filter.Len("name")}

// Not sure which one works best for Len + reference
f := filter.Cond{Target: filter.Len("bestFriend", "name")} // variadic
f := filter.Cond{Target: filter.Reference("bestFriend", filter.Len("name"))} // composition
```

In the latter example filter.Reference has to concatenate the target into some formatted string, which will add a tiny cost for then parsing the string in the `internal/api` layer, but that seems negligible.

bevzzz: > ```go
> // Not sure which one works best for Len + reference
> f := filter.Cond{Target: filter.Len("bestFriend", "name")} // variadic
> f := filter.Cond{Target: filter.Reference("bestFriend", filter.Len("name"))} // composition
> ```


Methinks the former, because the one with composition might give the idea that other targets are "composable", which they're not. Instead, the user has a simple choice:

```go
string
filter.Reference(... string) string
filter.ReferenceCount(... string) string
filter.Len(... string) string
```

WDYT?

tsmith023: If we can achieve the composition one I think that would be great from the PoV of doing filters through several references, e.g. `bestFriend.hasFriend.name`. The example given is for `bestFriend.len(name)` and I'm hoping that would work by composing reference inside it similarly: `filter.Cond{Target: filter.Reference("bestFriend", filter.Reference("hasFriend", "name"))}`

Agreed that the most common usage should be as simple as possible so having property just be a simple `string` makes sense. Also I like that the actual interface will still be `string` so that users can short-hand it if they so wish without using these helpers

internal/api/search.go
tsmith023: As always, the gRPC API may not have this validation at this stage 😞 So we need to check each time we add something here and decide whether to add it to the server (preferred) or just hot-fix the client for now

query/filter/filter.go
bevzzz: > filter.Reference("bestFriend", filter.Reference("hasFriend", "name"))

I don't know that I like that too much, because IMO that's more complex than just passing the "path" as strings. That shape reminded me of [the comment](https://github.com/golang/go/issues/61898#issuecomment-1672371746) regarding composition I read yesterday:

```
Multiple other languages have proven over and over that a left-to-right chained syntax is _vastly_ superior ergonomically to simple top-level functions [for iterators]
```

Perhaps I can do you one better:

```go
type Reference string

// Reference adds a reference "hop" to the path.
func (r Reference) Reference(string) Reference

// Property adds a property to the path and terminates it.
func (r Reference) Property(string) string

// Count applies count() operator to the last reference in the chain
func (r Reference) Count() string
```

Firstly, building deeply-nested targets becomes simple and intuitive:

```go
filter.Reference("bestFriend").Reference("hasFriends").Property("name")
```

Secondly, the signature of each function guides the user:

```go
filter.Reference("hasAwards").Count() // OK, count(reference) is valid
filter.Reference("hasAwards").Property("weight").Count() // Not possible, because Property returns string
```

Most importantly, we've decoupled "filter expression" from "filter target". Users who only need property-based targets are not forced to deal with the arcane method-chaining syntax and can continue using simple strings.

For cases where references are necessary, the syntax gets a bit more involved but leverages the type system to add some guardrails.

internal/api/search.go
bevzzz: > just hot-fix the client for now

My concern is these hotfixes end up as permanent residents, because deleting something from the codebase is much more difficult than adding it. Python and TS clients are living testaments to that and I find there (probably) aren't enough people using multi-reference target filters to justify the bloat these validations introduce. The users who really want to use these complex filters should spend some time to figure out the correct syntax for it IMO.

query/filter/filter.go
tsmith023: This would be perfect IMO as it achieves the fluent builder pattern that I think is natural for these types of filters. The fact that `Reference` is just an alias for `string` that you can still attach methods to is very nice thus allowing for the simpler string-only path syntax but also the fluent self-documenting syntax

tsmith023: Would we also have `func Property(string) string` to achieve symmetry between `filter.Reference(...).Property(...)` and `filter.Property(...)`? As a user, I'd expect to have both so that if I'm in the helper paradigm I stay there and don't ever provide raw strings

internal/api/search.go
tsmith023: For sure, can we document the specific missing validation on the server's issue board now and then have it fixed before v6 drops?

query/filter/filter.go
bevzzz: > Would we also have `func Property(string) string`

Absolutely not! The point is to _only_ involve the Reference helper (penalize) for the reference-property case and streamline the more common and simpler one. If all we need from the user is a simple string then that's exactly what they should be expected to pass.

> As a user, I'd expect to have both

I think you're expecting this as a developer who is keenly aware of the existence of both and, by nature of your position, is required to consider both in all scenarios.

As a user who's only interested in filtering by property (vast majority) there is little to gain from a purely cosmetic `func Property(s string) string { return s }` when you can just do `s`. Not to mention, that we'll be polluting the package's API.

In `query`, ReturnProperties and QueryProperties (bm25) take `[]string` and we don't provide any sugar like `QueryProperties: query.Properties("title", "description")`.

> to achieve symmetry between filter.Reference(...).Property(...) and filter.Property(...)

Remember that `Reference` is a type, not a function. It also exists for a very particular reason, namely, to hide the internals of what "reference target" string looks like and give us some flexibility there.

The `Len` function and the `Count` method exist for the same reason and are not cosmetic.

internal/api/search.go
bevzzz: I'll do that!

bevzzz: The new syntax for both single- and multi-target references should be significantly more error-prone.

```go
filter.Reference("hasFriends").Property("name") // single-target 
filter.Reference("hasAwards").Collection("GrammyAwards").Property("year") // multi-target
```

tsmith023: > should be significantly more error-prone

I hope you mean less 😅 

I'm happy to merge this now and release to get feedback but wdyt about this?
```go
filter.MultiReference("hasAwards", "GrammyAwards).Property("year")
```

tsmith023 (approved):

internal/api/search.go
bevzzz: > I hope you mean less

hahahah, definitely!
Somehow thought -prone meant "resilient" for a the moment I posted that.

funny, i'm sure i've used error-prone correctly a zillion times before

bevzzz: Right now I'm very much enjoying the fact that `Reference` provides a single entrypoint into working with references.

Let's merge and get feedback? Changing to `MultiReference` would be like a 5-line diff.

PR #384, opened 2026-04-26
feat: add sanity tests for BLOB_HASH property type

### LINKED ISSUE #377: Add support for blobHash property type
Release v1.37.0 introduces new `blobHash` property type, which saves in store blob hashes instead of the actual blob data.
This property type aims to reduce disk footprint when being used with multimodal modules.

antas-marcin:
## Summary
- Adds a `TestSchema_integration` subtest that creates a class with a `blobHash` property and verifies it round-trips through the schema API.

Closes #377

## Test plan
- [ ] CI integration tests pass against Weaviate `1.37.2`


go.mod
```diff
@@ -1,34 +1,34 @@
 module github.com/weaviate/weaviate-go-client/v5
 
-go 1.25
+go 1.26
```
bevzzz, line 3: Why do we need to update the required go version? Is it because of the `1.37.2` Weaviate import?

bevzzz (approved):

PR #386, opened 2026-04-27
v6: Move Insert to gRPC BatchObjects

bevzzz:
Traditionally our clients have exposed two distinct methods for writing data: `Insert` for single-object writes (REST-based) and `InsertMany` for (naive) batching (gRPC-based). In the `alpha.1` we also added the REST-based Insert, as it felt easier to implement.

Moving forward, we want to only use one transport mechanism for both 1 and N objects. From the developer's perspective it is easier to maintain. For the user it should not really matter, since transport is an implementation detail that's left up to the package authors.

This PR, then, makes `data.Client.Insert` variadic and uses gRPC `BatchObjects` request to send the data to the server. The method will no longer return a full object representation, which is in line with other clients (where we only return the object's ID) and, more importantly, batch-inserts, where the shape of the response is more or less dictated by the data which the server sends us back.

Effectively this PR also implements `InsertMany`.

__

Misc: fixed a flaky OIDC background refresh test.


internal/api/data.go
```diff
@@ -255,3 +222,79 @@ func (r *DeleteObjectRequest) Query() url.Values {
...
+func marshalProperties(properties map[string]any, dest *proto.BatchObject_Properties) error {
+	if len(properties) == 0 {
+		return nil
+	}
+	nonRef, err := structpb.NewStruct(properties)
+	if err != nil {
+		return err
+	}
+
+	// TODO(dyma): move object / array properties out of nonRef
```
bevzzz, line 265: I compared this to the Java implementation and noticed there we weren't placing object- and array- properties into the dedicated stub fields, e.g. `BooleanArrayProperties`. The server seems to be able to handle that, so I propose we enjoy the simpler implementation unless someone observes noticeable performance issues with it.

data/client.go
```diff
@@ -41,64 +42,83 @@ type Object struct {
...
+	Took   time.Duration
+	Errors map[uuid.UUID]string
+}
+
+// Insert new objects into the collection.
+//
+// To create objects without data or vectors and a UUID
+// generated by the server, pass nils as the second parameter.
+func (c *Client) Insert(ctx context.Context, objects ...*Object) (*InsertResult, error) {
+	batch := slices.Grow([]api.BatchObject(nil), len(objects))
```
bevzzz, line 55: This is a neat pattern I found equivalent to the snippet below:

```go
var batch []api.BatchObject
if len(objects) > 0 {
    batch = make([]api.BatchObject, len(objects))
}
```

I would like to avoid using `make` unconditionally, because for `len(objects) == 0` this produces an empty slice _which needs to be added to the expected value in unit tests_, polluting the latter. In contrast, slices.Grow preserves nil-ness.

tsmith023 (approved):

PR #394, opened 2026-05-04
v6: Add /live, /ready, and /meta utils

bevzzz:
This PR extends the top-level client with 3 methods for querying instance liveness, readiness, and metadata.


m3talux (approved):
LGTM!

PR #395, opened 2026-05-05
v6: Batch insert references

bevzzz:
This PR implements the public API for the `BatchReferences` RPC.
Crucially, we extend the `data.Reference` struct to contain the `Origin` field with the intent of using the same struct for reference properties as part of the inserted object and "standalone" references created via `.AddReferences()`.

A complete reference must contain two "object paths", the `origin` and the `target`. This is what you'll see in `internal/api/data.go`.

```go
type Reference { Source, Target ObjectPath}
```

To simplify the usage for the end user, we unpack Target's `Collection` and `UUID` fields in the "public" reference struct, such that the additional nesting doesn't get in the way:

```go
type Reference { 
  Origin ObjectPath; 
  /*Target*/Collection string; 
  /*Target*/UUID uuid.UUID;
}

// When used as a "reference property" (origin can be inferred from collection name + object ID + map key)
data.Object{
  UUID: uuid.New(),
  References: data.References{
    "bornIn": {UUID: "uuid-kansas"}, // single-target
    "marriedTo": {Collection: "Celebrities", UUID: "uuid-john-doe"}, // multi-target
  }
}

// When used to create a reference to an existing object (origin must be explicit):
data.Reference{
  Origin: data.ObjectPath{
      Collection: "Casino",
      Property: "robbedBy",
      UUID: "uuid-belagio",
  },
  Collection: "HollywoodRobbers",
  UUID: "uuid-dany-ocean"
}
```

The "origin" will also be necessary for inserting references via the SSB protocol, so this PR is a stepping stone towards that.


data/client.go
```diff
@@ -130,6 +139,48 @@ func (c *Client) Delete(ctx context.Context, id uuid.UUID) error {
 	return nil
 }
 
+type AddReferencesResult struct {
+	Took   time.Duration
+	Errors map[Reference]string
+}
+
+func (c *Client) AddReferences(ctx context.Context, references ...Reference) (*AddReferencesResult, error) {
```
tsmith023, line 147: Should this be in the singular, e.g. `AddReference`, if the objects alternative is just `Insert`?

tsmith023 (approved):

data/client.go
bevzzz: I considered that too, but I felt `AddReference(ctx, ref1, ref2, ref3)` is maybe awkward? It's easier w/ Insert, because it doesn't have a noun to pluralize.

I don't have a strong opinion about either, what do you think is preferable?

tsmith023: Hmm, I agree here. `AddReference(ctx, ref1, ref2, ref3)` is worse than `AddReferences(ctx, ref)`

PR #396, opened 2026-05-08
v6: Batch delete

bevzzz:
This PR adds support for `BatchDelete` operation, allowing users to delete objects which match the filtering criteria. 

Unlike in #386, I decided not to remove the single-target `Delete` method in favor of the batched solution. The input types are different, as Delete allows removing the object by just providing the UUID without fiddling with the filters. 

The batch delete method is called `DeleteSelected` but I'm happy to change that to `DeleteMany` or `BatchDelete`.

### DIFF
--- a/data/client.go
+++ b/data/client.go
@@ -11,6 +11,7 @@ import (
 	"github.com/weaviate/weaviate-go-client/v6/internal"
 	"github.com/weaviate/weaviate-go-client/v6/internal/api"
 	"github.com/weaviate/weaviate-go-client/v6/internal/dev"
+	"github.com/weaviate/weaviate-go-client/v6/query/filter"
 	"github.com/weaviate/weaviate-go-client/v6/types"
 )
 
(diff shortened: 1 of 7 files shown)


data/client.go
```diff
@@ -181,6 +182,40 @@ func (c *Client) AddReferences(ctx context.Context, references ...Reference) (*A
...
+	// DryRun reports UUIDs that matched the filter
+	// without deleting the objects.
+	DryRun bool
+
+	Verbose bool
+}
+
+type DeleteSelectedResult api.DeleteObjectsResponse
+
+func (c *Client) DeleteSelected(ctx context.Context, options DeleteSelected) (*DeleteSelectedResult, error) {
```
tsmith023, line 198: Why the divergence in naming here? Is it called `DeleteSelected` in the Java & C# clients? I think `DeleteMany` still makes sense despite the fact that you're deleting only those "many" that are "selected" by the filter, wdyt? Users that switch between langs (e.g. wcd devs) might be confused by the difference between this naming and that of the py/ts clients too

bevzzz: I started out with DeleteMany, but was a bit surprised to find out it didn't take "many IDs" the same way AddReferenceMany would take "many references", and InsertMany -- "many objects".

DeleteSelected is a bit more descriptive, which I like in this context.

I'm sure this won't survive the final rename-sweep like we had before Java GA, but do you think we can trial it during alpha.3? To see if my judgement is off here or not.

bevzzz: Consistent naming is very important too, so I definitely hear your concern about users switching between languages.

tsmith023: I'm happy to release like this and see what the feedback is as an experiment

bevzzz: Cool, let's do the experiment then

PR #397, opened 2026-05-12
v6(RBAC): Roles

bevzzz:
This PR adds support for a set of RBAC APIs related to role management and user/group assignment. It is available under `Client.Roles` namespace. Similarly to other client, permissions for the same resource are grouped under one object (unlike in the REST API), so that the AliasPermission looks like so:

```go
type AliasPermission struct{
    // Resource
    Collection, Alias string
    // Permitted actions
    Create, Read, Update, Delete bool
}
```

Consequently, the un-/marshaling in `internal/api/rbac.go` is quite verbose, but overall simple to follow (?), as the same pattern is repeated for every permission type / action. I did not want to resort to using reflection here and the only "heuristic" you'll find is the assumption that the last element of the "action" value in the REST body corresponds to the key for the resource. I.e `groups` in:

```json
{ "action": "assign_and_revoke_groups", "groups": { ... } }
```

### DIFF
--- a/internal/api/rbac.go
+++ b/internal/api/rbac.go
@@ -129,10 +129,11 @@ type (
 	UserPermission struct {
 		UserID string `json:"user,omitempty"`
 
-		Create bool `json:"-"`
-		Read   bool `json:"-"`
-		Update bool `json:"-"`
-		Delete bool `json:"-"`
+		Create          bool `json:"-"`
+		Read            bool `json:"-"`
+		Update          bool `json:"-"`
+		Delete          bool `json:"-"`
+		AssignAndRevoke bool `json:"-"`
 	}
 )
 
@@ -345,6 +346,8 @@ func (r *HasPermissionRequest) MarshalJSON() ([]byte, error) {
 		action, data = rest.UpdateUsers, r.Users
 	case r.Users.Delete:
 		action, data = rest.DeleteUsers, r.Users
+	case r.Users.AssignAndRevoke:
+		action, data = rest.AssignAndRevokeUsers, r.Users
 	}
 
 	permission := map[string]any{"action": action}
@@ -522,6 +525,9 @@ func (ps *Permissions) MarshalJSON() ([]byte, error) {
 		if p.Delete {
 			add(rest.DeleteUsers, p)
 		}
+		if p.AssignAndRevoke {
+			add(rest.AssignAndRevokeUsers, p)
+		}
 	}
 
 	return json.Marshal(permissions)
@@ -640,6 +646,8 @@ func (r *Role) UnmarshalJSON(data []byte) error {
 			find(lookup, p.Action, p.User, &r.Users).Update = true
 		case rest.DeleteUsers:
 			find(lookup, p.Action, p.User, &r.Users).Delete = true
+		case rest.AssignAndRevokeUsers:
+			find(lookup, p.Action, p.User, &r.Users).AssignAndRevoke = true
 		}
 	}
 
--- a/internal/api/rbac.go
+++ b/internal/api/rbac.go
@@ -45,10 +45,15 @@ func (r *CreateRoleRequest) Method() string { return http.MethodPost }
 func (r *CreateRoleRequest) Path() string   { return "/authz/roles" }
 func (r *CreateRoleRequest) Body() any      { return &r.Role }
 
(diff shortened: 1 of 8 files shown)


rbac/roles.go
```diff
@@ -0,0 +1,342 @@
...
+	}
+	return nil
+}
+
+type AddPermissions struct {
+	RoleID      string
+	Permissions Permissions
+}
+
+func (rc *RolesClient) AddPermissions(ctx context.Context, options AddPermissions) error {
```
bevzzz, line 138: I initially decided to make `RoleID` a field in the struct and not a standalone parameter to add some room for future changes, e.g. some other parameter identifying the role is added. Do you think that makes sense?

I'm not sure because `Delete()` above has `roleID string` and not a "options" struct. Thinking that Add/RemovePermissions should follow the same pattern.

internal/api/rbac.go
```diff
@@ -0,0 +1,695 @@
...
+
+		Create bool `json:"-"`
+		Read   bool `json:"-"`
+		Update bool `json:"-"`
+		Delete bool `json:"-"`
+	}
+	DataPermission struct {
+		Collection string `json:"collection,omitempty"`
+		Tenant     string `json:"tenant,omitempty"`
+		Object     string `json:"object,omitempty"`
```
tsmith023: I'm pretty sure object-level RBAC was deprecated until a definitive decision was made on exactly how it would look and work so we should probably remove this for now

rbac/roles.go
tsmith023: Happy to roll it like this and get feedback from usage. Breaking changes are okay during `alpha`

rbac/roles.go
```diff
@@ -0,0 +1,350 @@
...
+	Groups      GroupPermission
+	Nodes       NodesPermission
+	Replication ReplicationPermission
+	Roles       RolePermission
+	Tenants     TenantPermission
+	Users       UserPermission
+}
+
+// HasPermission checks if a role contains a permission.
+// Only one permission can be checked in a single call.
```
tsmith023, line 184: Given this, should `func (rc *RolesClient) HasPermission` perform this validation inside its body or is it expressly handled by the server?

internal/api/rbac.go
bevzzz: > we should probably remove this for now

Ack!

> I'm pretty sure object-level RBAC was deprecated

We need to update the [OpenaAPI schema](https://github.com/weaviate/weaviate/blob/e4d1259f4893852bd9fada3d3b52871353debd68/openapi-specs/schema.json#L183-L187)

rbac/roles.go
bevzzz: > is it expressly handled by the server

IMO these kind of validations should always be handled by the server, because it's the ultimate source of truth. I will check if that's actually the case.

bevzzz: Checked and [the server does no validation](https://github.com/weaviate/weaviate/blob/749aa12cb0d02920243cf641e115daf08a0f9820/adapters/handlers/rest/authz/handlers_authz.go#L332-L339) and always checks the first policy in the request.

Then I also realized that the Go client will only ever send a single permission even if multiple policies were set by the user, so the server wouldn't even know the request was bad. Which is fly I followed your suggestion and added a validation step in the client to make sure only 1 permission is being checked per request.

I tried to come up with a different shape of the input that would somehow prevent that but couldn't find anything satisfying.

rbac/roles.go
```diff
@@ -0,0 +1,496 @@
...
+
+// marshalHasPermission returns [api.HasPermissionRequest] with exacly 1 permission
+// (action + data) pair set and returns an error if more permissions are set in has.
+// This validation happens at this step, because internal/api expects the request
+// to be valid and will only send a single permission, even if more have been set,
+// so the server cannot perform this validation.
+//
+// NOTE(dyma): we might want to introduce some Validator interface in the transport
+// layer, which every request object can optionally implement, so that all validation
+// is done right before we marshal the request and dispatch to the actual transport.
```
tsmith023, line 348: I like this idea

tsmith023 (approved):

rbac/roles.go
bevzzz: I'll get to it the next time I need to write some validation code

PR #403, opened 2026-05-21
Send empty vector index type on Weaviate 1.37.5

mpartipilo:
Starting with Weaviate 1.37.5 the server sets a default vector index type to set when the client does not specify one. This PR establishes that no vector index type is set by default, and only sets "hsnw" as default for servers running <=1.37.4.

Ensure compatibility with Weaviate 1.37.5+ by sending an empty vector index type, allowing the server to apply the default. Add integration tests to verify behavior across versions and fix a bug related to mixing vector types. Introduce a helper function for environment variable overrides in tests.


bevzzz:
This change is not needed.

PR #404, opened 2026-05-24
v6(RBAC) Users / Groups

bevzzz:
This PR builds upon #397 and adds APIs for managing users and groups as well as their roles.
The API design mirrors that of the other clients and does not support the upcoming namespace feature yet. As such, it will likely need to be revisited later, when that feature stabilizes.

### DIFF
--- a/internal/api/rbac.go
+++ b/internal/api/rbac.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/weaviate/weaviate-go-client/v6/internal/api/internal/gen/rest"
+	"github.com/weaviate/weaviate-go-client/v6/internal/dev"
 	"github.com/weaviate/weaviate-go-client/v6/internal/transports"
 )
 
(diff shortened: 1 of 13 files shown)


internal/api/rbac.go
```diff
@@ -173,55 +173,39 @@ var GetAssignedUsersRequest = transports.IdentityEndpoint[string](http.MethodGet
...
 
-// RemovePermissionsRequest removes permissions from a role.
-type RemovePermissionsRequest struct {
+// ManagePermissionsRequest adds permissions to a role.
+type ManagePermissionsRequest struct {
 	transports.BaseEndpoint
 	RoleID      string      `json:"-"`
+	Action      string      `json:"-"`
 	Permissions Permissions `json:"permissions"`
 }
```
tsmith023, line 200: I think the renaming is good in principle but might there be confusion from a user between the meaning of `Action` within the contexts of both `ManagePermissionsRequest` and `Permissions`? wdyt?

bevzzz: Yes, agreed. I think the proper noun for that field would be `Verb`. (oxymoron intended)

I'll update before merging the PR.

PR #405, opened 2026-05-24
feat(tokenize): adapt to Weaviate 1.37.4 response shape changes

mpartipilo:
## Summary

Weaviate 1.37.4 ([weaviate/weaviate#11079](https://github.com/weaviate/weaviate/pull/11079)) changed the `/v1/tokenize` wire shape: response no longer echoes `tokenization` or `analyzerConfig`, and the request's `stopwordPresets` is now `[]string` instead of `map<string,StopwordConfig>` (inline custom presets → HTTP 400).

This PR keeps the existing SDK surface working across **1.37.0 – 1.37.4**. The SDK API rework needed to actually *use* custom presets on 1.37.4+ is a follow-up.

## Changes

- `weaviate/tokenize/text_tokenizer.go` — backfill `Tokenization` from the request when the server omits it.
- `weaviate/tokenize/property_tokenizer.go` — backfill via class-schema lookup; only paid when the server omitted the field.
- `test/testsuit/version.go` — add `ServerAtLeast` for tests that branch on server version; `AtLeastWeaviateVersion` wraps it.
- `test/tokenize/tokenize_test.go` — `Tokenization_Enum` asserts asymmetric query/indexed on 1.37.4+; `CustomPreset_*` and `AnalyzerConfig_Echoed` skip on 1.37.4+ with TODOs.
- `.github/workflows/workflow-tests.yaml` — `WEAVIATE_VERSION` 1.37.1 → 1.37.4.

## Test plan

- [x] `go test ./test/tokenize/...` against stock 1.37.0 and 1.37.4
- [ ] CI green at bumped version
- [ ] Follow-up: SDK rework for `stopwordPresets: []string`

### DIFF
--- a/test/tokenize/tokenize_test.go
+++ b/test/tokenize/tokenize_test.go
@@ -31,17 +31,21 @@ func TestTokenize_integration(t *testing.T) {
 	// -------- Serialization ------------------------------------------------
 
 	t.Run("Tokenization_Enum", func(t *testing.T) {
+		// Weaviate >=1.37.4 filters default stopwords on the query side for `word`;
+		// older versions return query == indexed.
+		queryStopwordsFiltered := testsuit.ServerAtLeast(t, client, "1.37.4")
 		cases := []struct {
-			name     string
-			method   tokenize.Tokenization
-			text     string
-			expected []string
+			name            string
+			method          tokenize.Tokenization
+			text            string
+			expectedIndexed []string
+			expectedQuery   []string // nil = same as expectedIndexed
 		}{
-			{"word", tokenize.Word, "The quick brown fox", []string{"the", "quick", "brown", "fox"}},
-			{"lowercase", tokenize.Lowercase, "Hello World Test", []string{"hello", "world", "test"}},
-			{"whitespace", tokenize.Whitespace, "Hello World Test", []string{"Hello", "World", "Test"}},
-			{"field", tokenize.Field, "  Hello World  ", []string{"Hello World"}},
-			{"trigram", tokenize.Trigram, "Hello", []string{"hel", "ell", "llo"}},
+			{"word", tokenize.Word, "The quick brown fox", []string{"the", "quick", "brown", "fox"}, []string{"quick", "brown", "fox"}},
+			{"lowercase", tokenize.Lowercase, "Hello World Test", []string{"hello", "world", "test"}, nil},
+			{"whitespace", tokenize.Whitespace, "Hello World Test", []string{"Hello", "World", "Test"}, nil},
+			{"field", tokenize.Field, "  Hello World  ", []string{"Hello World"}, nil},
+			{"trigram", tokenize.Trigram, "Hello", []string{"hel", "ell", "llo"}, nil},
 		}
 		for _, tc := range cases {
 			t.Run(tc.name, func(t *testing.T) {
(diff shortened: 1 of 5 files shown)


bevzzz:
@mpartipilo please remove the ~100 lines of commentrot from the diff and trim down the PR description to an adequate size.

mpartipilo:
> @mpartipilo please remove the ~100 lines of commentrot from the diff and trim down the PR description to an adequate size.

It should be a lot better now @bevzzz

bevzzz:
> > @mpartipilo please remove the ~100 lines of commentrot from the diff and trim down the PR description to an adequate size.
>
> It should be a lot better now @bevzzz

Please trim down the PR description to an adequate size and level of detail.

image

`*here`

mpartipilo:
@bevzzz PR update failed earlier. Now it's tight as it should be

test/tokenize/tokenize_test.go
```diff
@@ -124,6 +132,11 @@ func TestTokenize_integration(t *testing.T) {
 	})
 
 	t.Run("CustomPreset_Additions", func(t *testing.T) {
+		// 1.37.4 changed `stopwordPresets` to `[]string`; SDK still emits the
+		// legacy map. TODO: rework SDK API and re-enable (weaviate/weaviate#11079).
+		if testsuit.ServerAtLeast(t, client, "1.37.4") {
+			t.Skip("custom stopword presets need an SDK API rework for 1.37.4+ wire shape")
+		}
```
bevzzz, line 139: question: why doesn't this use the refactored `AtLeastWeaviateVersion`?

test/tokenize/tokenize_test.go
```diff
@@ -140,6 +153,9 @@ func TestTokenize_integration(t *testing.T) {
 	})
 
 	t.Run("CustomPreset_BaseAndRemovals", func(t *testing.T) {
+		if testsuit.ServerAtLeast(t, client, "1.37.4") {
```
bevzzz, line 156: question: why doesn't this use the refactored AtLeastWeaviateVersion?

test/tokenize/tokenize_test.go
```diff
@@ -169,6 +185,10 @@ func TestTokenize_integration(t *testing.T) {
 	})
 
 	t.Run("AnalyzerConfig_Echoed", func(t *testing.T) {
+		// Behaviour still covered by AsciiFold/StopwordPreset_String etc.
+		if testsuit.ServerAtLeast(t, client, "1.37.4") {
+			t.Skip("server no longer echoes AnalyzerConfig on response (weaviate/weaviate#11079)")
+		}
```
bevzzz, line 191: question: why doesn't this use the refactored AtLeastWeaviateVersion?

weaviate/tokenize/text_tokenizer.go
```diff
@@ -65,5 +65,9 @@ func (b *TextTokenizer) Do(ctx context.Context) (*TokenizeResult, error) {
 	if decodeErr := responseData.DecodeBodyIntoTarget(&result); decodeErr != nil {
 		return nil, decodeErr
 	}
+	// Weaviate >=1.37.4 no longer echoes `tokenization`; backfill from the request.
+	if result.Tokenization == "" {
+		result.Tokenization = b.tokenization
+	}
```
bevzzz, line 71: question: is this change even necessary? If the server decided it doesn't need to return "tokenization", why does the client need to cover for it? Do we know what was the reason for this field being dropped from the server response in the first place?

bevzzz (changes requested):

test/tokenize/tokenize_test.go
mpartipilo: One skips the test directly, the other just returns boolean to assert conditionally on server version

weaviate/tokenize/text_tokenizer.go
mpartipilo: CI started failing because of the change in the server. This is just a fix in the test so it keeps working and asserting in pre and post 1.37.4

test/tokenize/tokenize_test.go
mpartipilo: Ah I see it now. Ignore my previous comment. This is a leftover from the heavier change earlier. I'll close this PR. It will be superseeded by v6 soon anyway.

weaviate/tokenize/text_tokenizer.go
bevzzz: > This is just a fix in the test

The above is a change in the _production client code_, it adds an extra request every time a user tokenizes a piece of text

PR #407, opened 2026-05-28
feat(query): add OverAll query to do plain search

bevzzz:
This PR adds the OverAll query type, which allows the user to run a simple query without a semantic / keyword search component.

### DIFF
--- a/query/over_all.go
+++ b/query/over_all.go
@@ -1,0 +1,58 @@
+package query
+
+import (
+	"context"
+
+	"github.com/google/uuid"
+	"github.com/weaviate/weaviate-go-client/v6/internal"
+	"github.com/weaviate/weaviate-go-client/v6/internal/api"
+	"github.com/weaviate/weaviate-go-client/v6/query/filter"
+)
+
+type OverAll struct {
+	Limit                  int32            // Limit the number of results returned for the query.
+	Offset                 int32            // Skip the first N objects in the collection.
+	AutoLimit              int32            // Return objects in the first N similarity clusters.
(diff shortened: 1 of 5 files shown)


query/over_all.go
```diff
@@ -0,0 +1,58 @@
...
+import (
+	"context"
+
+	"github.com/google/uuid"
+	"github.com/weaviate/weaviate-go-client/v6/internal"
+	"github.com/weaviate/weaviate-go-client/v6/internal/api"
+	"github.com/weaviate/weaviate-go-client/v6/query/filter"
+)
+
+type OverAll struct {
```
tsmith023, line 12: I agree in principle that `query.OverAll` aligns with `aggregate.OverAll` but it is a divergence from the python and ts clients. Is it `OverAll` in Java/C# too? As with the `DeleteSelected` renaming, I'm happy if we merge it like this and release to get feedback

I have a suspicion that it might confuse platform/console/labs if this function isn't called `FetchObjects`, wdyt?

bevzzz: Fair question, and it is in divergence with Java/C# API, both of which have FetchObjects.

I do not particularly mind changing it to FetchObjects (I feel more strongly about DeleteSelected). Seeing how we don't have the capacity to run user-testing sessions though, I'd like to use the alpha stage to test out different "variations".
If there's pushback, I'm happy to rename before GA.

tsmith023: Agreed :D

tsmith023 (approved):

PR #408, opened 2026-05-29
v6: Objects Iterator

bevzzz:
This PR implements an object iterator compatible with Google's `iterator.Pager` following the guidelines published here: https://github.com/googleapis/google-cloud-go/wiki/Iterator-Guidelines.

The idea with using an existing pager implementation is to follow some wider-known pattern in the absence of an established iterator patter in Go std lib. Anyone familiar with Google Cloud Clients will find themselves at home; anyone new to the pattern can benefit from existing [documentation](https://github.com/googleapis/google-cloud-go/wiki/Iterator-Guidelines) and examples.

Following the naming convention, object iterator is exposed from a `collections.Handle` under the name `Objects`.

```go
songs := c.Collections.Use("Songs")
it := songs.Objects(ctx)

Play:
for {
    song, err := it.Next()
    switch err {
    case nil:
        play(song)
    case iterator.Done:
        break Play
    default:
        return err
    }
}
```

You can find more examples of working with the iterator in the guidelines doc linked above.

**Note:** it would be an interesting exercise to consider if we can have Pager work with `query.Decode` to support struct-properties. It is plain to do with the object returned from `Next`.

An important detail is that the `iterator` package has, presumably, been published before Go got generics, so Pager is not generic in the type of the objects it produces. Instead, it [relies on reflection to assign the result to its destination](https://github.com/googleapis/google-api-go-client/blob/8f434ff91fe8dc299942ecd7c87ebab151ff38e5/iterator/iterator.go#L207-L225). My understanding is that any performance overhead from using reflection is amortized over the size of the batch, i.e. we fetch 100 objects and assign them once, and then reflection is not used until the next batch is fetched. We can decide to fork / adapt the code from the `iterator` package in the future to avoid that, but I think it is a reasonable solution at the moment.

Reference RFC: https://app.notion.com/p/weaviate/RFC-Iterator-Pagination-35170562ccd6802aa8e9e840d7ff5683

### DIFF
--- a/query/iterator.go
+++ b/query/iterator.go
@@ -1,0 +1,65 @@
+package query
+
+import (
+	"context"
+
+	"github.com/google/uuid"
+	"google.golang.org/api/iterator"
+)
+
+func NewObjectIterator(ctx context.Context, c *Client) *ObjectIterator {
+	it := &ObjectIterator{
+		ctx: ctx,
+		c:   c,
+	}
+
+	it.pageInfo, it.nextFunc = iterator.NewPageInfo(
+		it.fetch,
+		func() int { return len(it.items) },
+		func() any { b := it.items; it.items = nil; return b })
+
+	return it
+}
+
+type ObjectIterator struct {
+	ctx      context.Context
+	c        *Client
+	pageInfo *iterator.PageInfo
+	nextFunc func() error
+	items    []*Object[map[string]any]
+}
+
+var _ iterator.Pageable = (*ObjectIterator)(nil)
+
+func (it *ObjectIterator) PageInfo() *iterator.PageInfo { return it.pageInfo }
+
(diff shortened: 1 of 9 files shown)


query/iterator.go
```diff
@@ -0,0 +1,71 @@
...
+
+	return it
+}
+
+type ObjectIterator struct {
+	ctx      context.Context
+	c        *Client
+	pageInfo *iterator.PageInfo
+	nextFunc func() error
+	items    []*Object[map[string]any]
```
tsmith023, line 43: Users will still be able to call query.Decode on these objects, right?

tsmith023 (approved):

query/iterator.go
bevzzz: Sure, these are the same objects that the queries return

PR #425, opened 2026-06-30
v6: Aliases

bevzzz:
This PR adds a CRUD client for `/aliases` endpoints.

```go
c.Alias.Create(ctx, alias.Alias{Collection: "Alex", Alias: "A"})
c.Alias.Get(ctx, "A")
c.Alias.List(ctx)
c.Alias.Update(ctx, alias.Alias{Alias: "A", Collection: "Anna"})
c.Alias.Delete(ctx, "A")
```

### DIFF
--- a/alias/client_test.go
+++ b/alias/client_test.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/stretchr/testify/require"
 	"github.com/weaviate/weaviate-go-client/v6/alias"
+	"github.com/weaviate/weaviate-go-client/v6/collections"
 	"github.com/weaviate/weaviate-go-client/v6/internal/api"
 	"github.com/weaviate/weaviate-go-client/v6/internal/testkit"
 )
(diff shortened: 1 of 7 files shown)


alias/client_test.go
```diff
@@ -0,0 +1,230 @@
...
+func TestNewClient(t *testing.T) {
+	require.Panics(t, func() {
+		alias.NewClient(nil)
+	}, "nil transport")
+}
+
+func TestClient_Create(t *testing.T) {
+	for _, tt := range []struct {
+		name  string
+		alias alias.Alias
```
bevzzz:
```suggestion
		alias collections.Alias
```

Should we export Alias type from the `collections` package to avoid stutter in the type name? I'm leaning yes, but I don't know if it's not a bit awkward.

tsmith023: A user defining an `alias` variable would collide with the package name, right? I.e., would the following be forbidden?

```go
import "github.com/weaviate/weaviate-go-client/v6/alias"

alias := alias.Alias{ ... }
```

if so I think re-exporting from `collections` is the least worst option. wdyt?

tsmith023 (approved):
Now we see the full power of your transport abstraction. This adds the full aliases functionality and doesn't have to even think about building and parsing http requests and responses (outside of meeting an interface, which itself is client-specific in its definition) 🚀

bevzzz:
> Now we see the full power of your transport abstraction

I'm also very much enjoying the ergonomics of having different Endpoint helpers like `StaticEndpoint` and `IdentityEndpoint`. The whole PR took about an hour to materialize, with tests and all.

alias/client_test.go
bevzzz: So `alias := alias.Alias{}` would be legal but only once in a scope, afterwards `alias` the package gets shadowed by the variable.

I like your point that exporting `collections.Alias` leaves the "alias" identifier for the user code! And I think the it reads much better.

PR #427, opened 2026-07-01
v6: Cluster API

bevzzz:
This PR adds a client for listing nodes and shards in a cluster.

```go
shards, err := c.Cluster.ListShardReplicas(ctx, cluster.ListShardReplicasOptions{
  Collection: "Songs",
  Shard: "xyz", 
})

nodes, err := c.Cluster.ListNodes(ctx, cluster.ListNodesReplicasOptions{
  Collection: "Songs",
  Shard: "xyz",
  Verbose: true,
})
```

### DIFF
--- a/cluster/client.go
+++ b/cluster/client.go
@@ -24,7 +24,7 @@ type ShardReplica api.ShardReplica
 
 type ListShardReplicasOptions struct {
 	Collection string // Required parameter.
-	Shard      string // Optional: if ommitted, all shards are returned.
+	Shard      string // Optional: if omitted, all shards are returned.
 }
 
 // ListShardReplicas returns a list of shards and all their replicas.
@@ -36,7 +36,7 @@ func (c *Client) ListShardReplicas(ctx context.Context, options ListShardReplica
 
 	var resp api.GetShardsResponse
 	if err := c.transport.Do(ctx, req, &resp); err != nil {
-		return nil, fmt.Errorf("get sharding state: %w", err)
+		return nil, fmt.Errorf("list shard replicas: %w", err)
 	}
 
 	out := make([]ShardReplica, len(resp))
@@ -98,7 +98,7 @@ const (
 )
 
 type ListNodesOptions struct {
-	Collection string // Optional: if ommitted, all nodes are returned.
+	Collection string // Optional: if omitted, all nodes are returned.
 	Shard      string // Optional: if set, only nodes containing this shard are returned.
 	Verbose    bool   // Optional: set to true to fetch complete node info.
 }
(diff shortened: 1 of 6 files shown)


cluster/client.go
```diff
@@ -0,0 +1,156 @@
...
+type Client struct {
+	transport internal.Transport
+}
+
+type ShardReplica api.ShardReplica
+
+type ListShardReplicasOptions struct {
+	Collection string // Required parameter.
+	Shard      string // Optional: if ommitted, all shards are returned.
+}
```
bevzzz, line 28: note: Instead of going for a fancier split between how optional and required params are presented, I went for a more functional option of the "options" struct.

The more I'm working on this client the more I'm leaning towards (ctx + options) shape for _most_ of the client's APIs.

WDYT?

cluster/client.go
```diff
@@ -0,0 +1,156 @@
...
+
+// ListShardReplicas returns a list of shards and all their replicas.
+func (c *Client) ListShardReplicas(ctx context.Context, options ListShardReplicasOptions) ([]ShardReplica, error) {
```
bevzzz, line 31: note: other clients return a "sharding state" struct, which has the Collection field in addition to the list of shards. I decided not to define another data type only to include a value which the user's already passed as an argument (collection name).

WDYT?

cluster/client.go
tsmith023: I think it makes sense as it's more defensive for the future

cluster/client.go
```diff
@@ -0,0 +1,156 @@
...
+	req := &api.GetShardsRequest{
+		Collection: options.Collection,
+		Shard:      options.Shard,
+	}
+
+	var resp api.GetShardsResponse
+	if err := c.transport.Do(ctx, req, &resp); err != nil {
+		return nil, fmt.Errorf("get sharding state: %w", err)
```
tsmith023: Given the immediate above, should this instead be `list sharding state: %w`?

cluster/client.go
```diff
@@ -0,0 +1,156 @@
...
+
+const (
+	NodeModeReadOnly  = NodeMode(api.NodeModeReadOnly)
+	NodeModeReadWrite = NodeMode(api.NodeModeReadWrite)
+	NodeModeScaleOut  = NodeMode(api.NodeModeScaleOut)
+	NodeModeWriteOnly = NodeMode(api.NodeModeWriteOnly)
+)
+
+type ListNodesOptions struct {
+	Collection string // Optional: if ommitted, all nodes are returned.
```
tsmith023: omitted 😜 (also in `ListShardReplicasOptions`)

cluster/client.go
```diff
@@ -0,0 +1,156 @@
...
+		Verbosity:  api.NodeVerbosityMinimal,
+	}
+
+	if options.Verbose {
+		req.Verbosity = api.NodeVerbosityVerbose
+	}
+
+	var resp api.GetNodesResponse
+	if err := c.transport.Do(ctx, req, &resp); err != nil {
+		return nil, fmt.Errorf("get sharding state: %w", err)
```
tsmith023: Should this instead be `"list nodes: %w"`

tsmith023 (approved):

PR #429, opened 2026-07-08
v6: Vectorizer modules

bevzzz:
This PR extends collection configuration to include vectorizers. It also provides one vectorizer module, `text2vec-module2vec`, as an example of the pattern in action.

We include support for "none" vectorizer, or self-provided vectors, which is achieved by simply not setting a vectorizer.

```go
c.Collections.Create(ctx, collections.Collection{
        // Default vector index + text2vec-model2vec vectorizer
        "lyrics_vec": {
            Vectorizer: model2vec.Text2Vec{
                URL: "example.com",
                Properties: []string{"lyrics", "meaning"},
            },
        },
        // "none" vectorizer (+ default vector index) will be configured
        "title_vec": {}, 
    },
})
```

The fact that Vectorizer accepts a `modules.Module` interface means we can easily accommodate "custom" modules both on read and write paths. E.g., suppose a new `text2vec-gzip` module is available in Weaviate, but the Go client hasn't been updated yet. The user can implement their own module and use it with the client.

```go
package my_gzip

import "github.com/weaviate/weaviate-go-client/v6/modules"

func init() { modules.Register(*new(Text2Vec)) }

type Text2Vec struct {
    CompressionLevel int `json:"compression_level"`
    Properties []string  `json:"properties"`
}

func (Text2Vec) Name() string { return "text2vec-gzip" }
```

The client can also read modules which it doesn't know about into a catch-all CustomModule:

```go
config, _ := songs.Config.Get(ctx) // Config.Get is not implemented yet
lyricsVec := config.Vectors["lyrics"]

// Suppose we have created the `text2vec-gzip` vector in a different app altogether,
// so this instance of Go client doesn't know about my_gzip.Text2Vec.
custom, ok := lyricsVec.Vectorizer.(modules.Custom)
if !ok {
    log.Fatalf("Not a modules.Custom, this is actually %T", lyricsVec.Vectorizer)
}

log.Print(custom.Name()) // Outputs: text2vec-gzip
log.Print(custom.Raw())  // Outputs: map[string]any{ "compression_level": 9 }
```

Here I'm trying out the "modules" pattern, where every 3rd party provider (openai/digitalocean/model2vec) declares their modules (vectorizer, reranker, etc) in a "shared" `modules` package, so that the user can easily discover different modules from the same provider.

E.g. `cohere.` + Tab will show all possible modules Cohere has.

___

I did a small change in `internal.Encode` such that it no longer requires a generic parameter like Decode does. The actual type is irrelevant for encoding.

### DIFF
--- a/collections/client_test.go
+++ b/collections/client_test.go
@@ -126,6 +126,14 @@ func TestClient_Create(t *testing.T) {
 						Collections: []string{"Singers", "Bands"},
 					},
 				},
+				Vectors: map[string]collections.VectorConfig{
+					"lyrics_vec": {},
+					"title_vec": {
+						Vectorizer: testkit.Module{
+							"url": "example.com",
+						},
+					},
+				},
 				Sharding: &collections.ShardingConfig{
 					DesiredCount:        3,
 					DesiredVirtualCount: 150,
@@ -206,6 +214,14 @@ func TestClient_Create(t *testing.T) {
 								{
 									Name:        "artist",
 									Collections: []string{"Singers", "Bands"},
+								},
+							},
+							Vectors: map[string]api.VectorConfig{
+								"lyrics_vec": {},
+								"title_vec": {
+									Vectorizer: testkit.Module{
+										"url": "example.com",
+									},
 								},
 							},
 							Sharding: &api.ShardingConfig{
@@ -355,6 +371,14 @@ func TestClient_GetConfig(t *testing.T) {
 								Collections: []string{"Singers", "Bands"},
 							},
 						},
+						Vectors: map[string]api.VectorConfig{
+							"lyrics_vec": {},
+							"title_vec": {
+								Vectorizer: testkit.Module{
+									"url": "example.com",
+								},
+							},
+						},
 						Sharding: &api.ShardingConfig{
 							DesiredCount:        3,
 							DesiredVirtualCount: 150,
@@ -434,6 +458,14 @@ func TestClient_GetConfig(t *testing.T) {
 					{
 						Name:        "artist",
 						Collections: []string{"Singers", "Bands"},
+					},
+				},
+				Vectors: map[string]collections.VectorConfig{
+					"lyrics_vec": {},
+					"title_vec": {
+						Vectorizer: testkit.Module{
+							"url": "example.com",
+						},
 					},
 				},
 				Sharding: &collections.ShardingConfig{
--- a/collections/client_test.go
+++ b/collections/client_test.go
@@ -217,10 +217,15 @@ func TestClient_Create(t *testing.T) {
 								},
 							},
 							Vectors: map[string]api.VectorConfig{
-								"lyrics_vec": {},
+								"lyrics_vec": {
+									Vectorizer: api.NoneVectorizer,
+								},
 								"title_vec": {
-									Vectorizer: testkit.Module{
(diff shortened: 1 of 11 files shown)


collections/client_test.go
```diff
@@ -126,6 +126,14 @@ func TestClient_Create(t *testing.T) {
 						Collections: []string{"Singers", "Bands"},
 					},
 				},
+				Vectors: map[string]collections.VectorConfig{
+					"lyrics_vec": {},
```
tsmith023, line 130: I think this is the incorrect UX given what is assumed by sending a blank vectoriser to weaviate

Sending nothing as the vectoriser should populate the vectoriser from the server-side defined `DEFAULT_VECTORIZER_MODULE` env var rather than the client implicitly assuming nothing to mean `self_provided`

tsmith023 (commented):
wdyt about adding an example test making use of the introduced `model2vec` vectoriser showing its usage end-to-end? I.e., also then reading the configuration back from weaviate in a `GET v1/schema/{collection}` call (if possible)?

collections/client_test.go
```diff
@@ -126,6 +126,14 @@ func TestClient_Create(t *testing.T) {
...
+					"title_vec": {
+						Vectorizer: testkit.Module{
+							"url": "example.com",
+						},
+					},
```
tsmith023, line 135: Actually, rather than introducing the e2e examples journey using `model2vec`, can we use the `model2vec` module here instead to confirm its usage?

collections/client_test.go
bevzzz: Ah, of course! You're totally right, I forgot that server-side defaults were a thing.
The `nil` vectorizer should have those semantics, not those of a "none" vectorizer.

collections/client_test.go
bevzzz: > can we use the model2vec module here instead to confirm its usage?

We could, but then we'd need to extend this test suite with test cases for every module we add down the road. I'd prefer to use a mock Module here (to see that _any abstract module can work_) and add a smaller and simpler test suite to `modules` package to check that each module produces the expected map/JSON config.

Beyond that, i.e. how that config is interpreted and use, it is the server's realm of responsibility.

IM0

> rather than introducing the e2e examples

I would use `model2vec` in the e2e suite because (I believe) it's what our test WCD cluster has configured.

collections/client_test.go
bevzzz: I've refactored some code around modules in the follow-up vector index PR, so it was easier to introduce that change there (no merge conflicts). Is it fine by you if we merge this PR as is?

See [this commit](https://github.com/weaviate/weaviate-go-client/pull/431/commits/61ef3de5631b89854ad6315a2b1b08975c66f980).

collections/client_test.go
bevzzz: I added a dedicated test suite for checking module configurations to make sure the tags are setup correctly and all known modules are registered.

Same as above, I did it in the #431 rather than here because that PR has an updated layout of the `modules` registry plus the `selfprovided.Vectorizer` module.

See [this commit](https://github.com/weaviate/weaviate-go-client/pull/431/commits/bc50a2701a314f5cd401f7db59a7c29b39533f42).

tsmith023 (approved):

PR #431, opened 2026-07-09
v6: Vector index configuration

bevzzz:
This PR introduces vector index configuration to the collection's vectors. Building on the "module" abstraction from #429 and leveraging `internal.Modules` registry, we define "vector index" to be a special module type.

The new `collections/vectorindex` package exports configuration objects for HFresh and (in the follow-up PRs) other index types. By giving `internal.Module` a generic parameter we achieve something like a tagged union, which we can re-use in different parts of the collection config by defining custom "tag" types.

```go
// Third-party modules like vectorizers and rerankers
type Module internal.Module[string]

// VectorIndex modules, i.e. hfresh, hnsw, ...
type VectorIndex internal.Module[vectorindex.Type]

// Compression modules, i.e. bq, pq, sq, ...
type Compression internal.Module[compression.Type]
```

The zero-value semantics for all three of them makes it quite easy to use the "default" option in the configuration:

```go
collections.Collection{
        "lyrics_vec": {
            Index: vectorindex.HFresh{ SearchProbe: 256 },
            /* Vectorizer: nil, */ (omitted, so the client will send "none": {})
        },
    },
}
```

___

I also considered using a struct-union pattern, like we have for distance/certainty, but decided against it. This is convenient for passing parameters (write) but not for reading the information back -- you'd need to either make the struct fields public or provide some getters for it.

```go
type Similarity { distance, certainty *float64 }
func Distance(d float64) *Similarity
func Certainty(c float64) *Similarity
```


modules/register.go
```diff
@@ -16,23 +16,14 @@ import (
...
-// Encode creates a map of all public fields of the module struct.
-func Encode(m Module) (map[string]any, error) {
-	return registry.Encode(m)
-}
-
-// Decode creates a new instance of the module struct.
-// The type depends on the type registered for the name.
-func Decode(name string, raw map[string]any) (Module, error) {
-	return registry.Decode(name, raw)
+	Registry.Register(m)
```
bevzzz, line 28: I decided to export the Registry itself instead of echoing its Encode/Decode methods for the sake of making the API surface of the `modules` package smaller.

That means that the user can, in theory, replace the Registry with another one, but I can't imagine why they'd want to do that really.

collections/vectorindex/index.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"github.com/weaviate/weaviate-go-client/v6/internal"
+)
+
+// Registry stores all vector index types defined by this package.
+var Registry internal.Modules[Type]
+
+func init() {
+	Registry.Register(*new(HFresh))
+}
```
bevzzz, line 12: Do you think we should put this registry behind `collections/internal` so that it wouldn't be accessible? It's only really meant to be used by the `collections` and `vectorindex` packages, but I've put it here because I didn't want to create a new package for just one value.

Happy to change that.

collections/vectorindex/index.go
```diff
@@ -0,0 +1,36 @@
...
+
+type Type string
```
tsmith023, line 14: Given the `DistanceX` enums below, should we have enums here for vector index types?

Then we'd have below:
```go
func (HFresh) Name() Type { return TypeHFresh } 
```
wdyt?

modules/register.go
tsmith023: Agreed, simplification here is a good idea imo

collections/vectorindex/index.go
tsmith023: I'd maintain the convention that has been established by the `v6-vectorizer` branch despite the extra packaging to keep the separation clear. Then all things registry clearly live with a specific pattern and don't accidentally leak into other areas of the codebase

collections/vectorindex/index.go
bevzzz: > Given the DistanceX enums below

This is a good illustration for what I wrote [here](https://github.com/weaviate/weaviate-go-client/pull/432#discussion_r3571815256) wrt compression types. Namely, that Distance is a value the user needs to provide, which is why the package enumerates the possible options. In contrast, `"hfresh"` can be treated as an "opaque" ID for the `HFresh` module.

bevzzz: We declare a new `type Type string` to limit what users can pass to `VectorConfig.Index` and avoid confusion when creating VectorConfig that takes 3 arguments which are all "modules".

bevzzz:
> I'd maintain the convention that has been established by the v6-vectorizer branch despite the extra packaging to keep the separation clear. Then all things registry clearly live with a specific pattern and don't accidentally leak into other areas of the codebase

I'm not sure which convention you have in mind. ~~Do you mean one where a package only exposes a `Registry()` hook but not the registry itself?~~ (I realised that `modules` does export the registry)

UPD: I just saw this same comment in [this thread](https://github.com/weaviate/weaviate-go-client/pull/431#discussion_r3570102434) with more context. I'll answer there.

collections/vectorindex/index.go
bevzzz: To clarify -- you'd prefer a `collections/internal` that has both a vector index and compression (after #432) registries in them, so that the users do not see `vectorindex.Registry` when working with the package?

tsmith023: Sorry for ambiguity/confusion in my reply. My preference would be for a chosen convention to be stuck to so if the `modules` registry is publicly available then other registries can be too and vice versa. Is that possible and does it add/reduce the boilerplate overhead?

bevzzz: OK, I think I understand now.

> if the modules registry is publicly available then other registries can be too

Here, `vectorindex.Registry` is publicly available, the only difference being that the package is located at `collections/vectorindex` whereas modules are at root level.

Same is true for `collections/compression` in #432.

tsmith023: Ah okay, so since `modules.Registry` is also public (I mistakenly assumed it wasn't) then having these be public is good imo

tsmith023 (approved):

PR #432, opened 2026-07-10
v6: Compression

bevzzz:
In this PR we complete the vector configuration with a compression algorithm. Compression modules are organized exactly like vectorizer and vector index modules from the previous PRs, i.e. under a dedicated `collections/compression` package. 

```go
collections.Collection{
        "lyrics_vec": {
            Compression: compression.RQ{
                Enabled: true,
                Bits: 8,
                RescoreLimit: 64,
                Cache: true,
            },
        },
    },
}
```

We start with only providing RQ quantization; a follow-up PR will extend the package with the remaining option (PQ/SQ/BQ).

### DIFF
--- a/collections/compression/compression.go
+++ b/collections/compression/compression.go
@@ -14,7 +14,6 @@ type Type string
 var _ internal.Module[Type] = (*RQ)(nil)
 
 type RQ struct {
-	Enabled      bool `json:"enabled"`
 	Bits         int  `json:"bits"`
 	RescoreLimit int  `json:"rescore_limit"`
 	Cache        bool `json:"cache"`
--- a/collections/compression/compression.go
+++ b/collections/compression/compression.go
@@ -1,0 +1,23 @@
+package compression
+
+import "github.com/weaviate/weaviate-go-client/v6/internal"
+
+// Registry stores all compression algorithms defined by this package.
+var Registry internal.Modules[Type]
+
+func init() {
+	Registry.Register(*new(RQ))
+}
+
+type Type string
+
+var _ internal.Module[Type] = (*RQ)(nil)
+
+type RQ struct {
+	Enabled      bool `json:"enabled"`
+	Bits         int  `json:"bits"`
+	RescoreLimit int  `json:"rescore_limit"`
+	Cache        bool `json:"cache"`
+}
+
+func (RQ) Name() Type { return "rq" }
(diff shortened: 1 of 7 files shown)


collections/compression/compression.go
```diff
@@ -0,0 +1,23 @@
...
+func init() {
+	Registry.Register(*new(RQ))
+}
+
+type Type string
+
+var _ internal.Module[Type] = (*RQ)(nil)
+
+type RQ struct {
+	Enabled      bool `json:"enabled"`
```
bevzzz: question: Do you think it's worth making it so that Enabled is always set to `true` when the user selects any compression, such that the following is enough to enable it?

```go
collections.VectorConfig{ Compression: compression.RQ{Bits: 8}}
```

tsmith023: Yes, I think so. The other clients don't even expose the ability to set `Enabled: false` and rather the presence of the config option is implicitly understood to mean "enabled"

tsmith023: The server still needs to receive `Enabled: true` to turn the compression on but, from the PoV of the user, it's needless information imo

collections/compression/compression.go
```diff
@@ -0,0 +1,22 @@
...
+import "github.com/weaviate/weaviate-go-client/v6/internal"
+
+// Registry stores all compression algorithms defined by this package.
+var Registry internal.Modules[Type]
+
+func init() {
+	Registry.Register(*new(RQ))
+}
+
+type Type string
```
tsmith023, line 12: Likewise for vector indices, should we define `TypePQ`, etc. here that are returned by `func (PQ) Name() Type { return TypeRQ }` functions below?

internal/api/collections.go
```diff
@@ -377,11 +396,32 @@ func (c *Collection) UnmarshalJSON(data []byte) error {
 			break
 		}
 
-		if v.VectorIndexConfig != nil {
+		if conf := v.VectorIndexConfig; conf != nil {
+			// Compression config map is embedded into the vector index configuration.
+			// If we can find any of the registered compresison modules there, we can
```
tsmith023: `compresison` typo 😜

bevzzz: Thank you, fixed!

collections/compression/compression.go
bevzzz: I think not. The qualifier (module ID) returned from Name is intentionally opaque, because it's not part of the package's public contract. Primarily, because other packages never need to use `compression.TypePQ` anywhere.

The compiler probably hoists this value into a constant so we aren't paying a penalty.

tsmith023 (approved):

PR #434, opened 2026-07-17
v6: Server-Side Batching

bevzzz:
This PR adds a client implementation for the SSB protocol. The functionality is accessible via collection handle's `Batch(ctx)` method.

The core logic is almost entirely contained in `internal/api/ssb/protocol.go` with the exception of Start / Stop messages, which the transport sends when it opens / closes the stream respectively. The rest of the changes are mostly "plumbing" the underlying gRPC client to the SSB client and un-/marshaling the messages.

Not included in this PR are also tests for the protocol implementation. #437 adds an "example" test for that happy-path and a more comprehensive (fuzz-) testing will come in a follow-up PR.

---

OK, now that we're done with all formalities, let's look at the API!
It's delightfully minimal:

```go
songs := c.Collections.Use("Songs")

b := songs.Batch(ctx) // 1) Open the batch

var tasks []*batch.Task
for o := range batchData {
  t, err := b.Object(o) // 2) Add data (b.Object or b.Reference)
  if err != nil {
    break // A non-nil error is always context.Canceled
  }
}
if err := b.Close(); err != nil { // 3) Close the batch
    log.Printf("batch stream exited abnormally: %v", err)
}

for _, t := range tasks {
    if err := t.Wait(); err != nil { // 4) Collect results for individual tasks
        log.Printf("task %s failed: %v", t.ID(), err)
    }
}
```


collections/client.go
```diff
@@ -90,6 +92,27 @@ func (h *Handle) Objects(ctx context.Context) *query.ObjectIterator {
...
 
+// Count objects in the collection, respecting the tenant if provided.
+func (h *Handle) Count(ctx context.Context) (int64, error) {
+	req := api.CountObjectsRequest(h.defaults)
+	var resp api.CountObjectsResponse
+	if err := h.transport.Do(ctx, &req, &resp); err != nil {
+		return 0, fmt.Errorf("count objects: %w", err)
+	}
+	return resp.Int64(), nil
+}
```
bevzzz, line 102: Moved this up so that it's grouped with the `Objects` iterator.

collections/client.go
```diff
@@ -90,6 +92,27 @@ func (h *Handle) Objects(ctx context.Context) *query.ObjectIterator {
...
+
+// Batch opens a new batch stream. The context will be used throughout
+// the whole streaming process and may be used to terminate it abruptly.
+// In a normal course of operation, a batch should be closed explicitly.
+func (h *Handle) Batch(ctx context.Context, options ...batch.Option) (*batch.Client, error) {
```
bevzzz, line 107: question: Should context be optional? We can always fallback to `context.Backgrond`. I'm leaning yes.

data/client.go
```diff
@@ -64,22 +64,7 @@ type InsertResult struct {
...
-		} else {
-			id = uuid.New()
-		}
-		batch = append(batch, api.BatchObject{
-			UUID:       id,
-			Properties: o.Properties,
-			Vectors:    apiVectors(o.Vectors),
-			References: apiReferences(o.References),
-		})
+		batch = append(batch, o.ToAPI())
```
bevzzz, line 67: Moved this bit to a method to reuse in the `batch` package.

data/client.go
```diff
@@ -149,13 +134,7 @@ func (c *Client) AddReferences(ctx context.Context, references ...Reference) (*A
...
 
 	for _, ref := range references {
-		batch = append(batch, api.Reference{
-			Origin: api.ObjectPath(ref.Origin),
-			Target: api.ObjectPath{
-				Collection: ref.Collection,
-				UUID:       ref.UUID,
-			},
-		})
+		batch = append(batch, ref.ToAPI())
```
bevzzz, line 137: Moved this bit to a method to reuse in the batch package.

data/client.go
```diff
@@ -245,3 +224,33 @@ func apiReferences(rs References) api.References {
 	}
 	return out
 }
+
+func (o *Object) ToAPI() api.BatchObject {
```
tsmith023, line 228: Is this method expected to be used by our users or is it purely for internal usage? If the latter I'd make it private to hide it away and avoid exposing `api.BatchObject` publicly. wdyt? (likewise for `Reference.ToApi` below)

batch/client.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+}
+
+type Client struct {
+	protocol *ssb.Client
+}
+
+func (c *Client) Object(o *data.Object) (*Task, error) {
+	data := o.ToAPI()
+	return c.add(ssb.Data{Object: &data})
```
tsmith023: wdyt about defining a constructor for an object (and reference below) to avoid raw struct initialisation outside of the package? Something like:
```go
ssb.NewObject(o.toAPI())
ssb.NewReference(r.toAPI())
```
imo, the abstraction is cleaner with private structs that are created using public init methods but that is an opinionated opinion so would love your thoughts too!

collections/client.go
tsmith023: I reckon it's a good idea to have it explicitly required here. Streaming will be a long-lived process so encouraging users to actively declare `ctx` will help them to think of timeouts/cancellation

collections/client.go
```diff
@@ -90,6 +92,27 @@ func (h *Handle) Objects(ctx context.Context) *query.ObjectIterator {
...
+	st, ok := h.transport.(stream.Transport)
```
tsmith023: This requirement is somewhat confusion and potentially smelly to me. Why should we need to perform a type assertion on an interface here rather than expect for the necessary streaming methods to be available on `h.transport` (of type `internal.Transport`)?

What is the benefit of defining `stream.Transport` separately to `internal.Transport` that necessitates the confusing interface behaviour here?

collections/client.go
bevzzz: We'll keep this a required parameter.

internal/api/ssb/protocol.go
```diff
@@ -0,0 +1,648 @@
...
+}
+
+// BatchRequest is a sized container, which accumulates batch items
+// until their size reaches the parent stream's capacity.
+type BatchRequest interface {
+	Add(v any) (added, full bool)
+}
+
+// RetryFunc determines if a task should be retried after the following error.
+type RetryFunc func(id string, retries int, err error) bool
```
tsmith023: To make it explicit this function is a predicate, wdyt about renaming it to `ShouldRetryFunc`? When I glanced at `RetryFunc`, I thought it might have an action/side-effect of itself

internal/api/ssb/protocol.go
```diff
@@ -0,0 +1,656 @@
...
+			case <-c.ctx.Done():
+				return
+			}
+		}
+	}()
+}
+
+// send consumes queue and retry channels, prepares and sends the batch.
+func (c *Client) send(ctx context.Context, s Stream) {
+	defer s.Close() //nolint:errcheck
```
tsmith023, line 332: Should we log the error here rather than discarding it?

internal/api/ssb/protocol.go
```diff
@@ -0,0 +1,656 @@
...
+			}
+
+			t.setValue(v)
+			c.wip.put(t)
+			c.batch.add(v)
+
+		case tasks := <-c.retry:
+			for _, t := range tasks {
+				c.batch.add(t.value())
+			}
```
tsmith023: Might it be worthwhile here to have `c.batch.addMany` to avoid hot locking-unlocking of `b.mu` in the tight loop?

internal/api/ssb/protocol.go
```diff
@@ -0,0 +1,656 @@
...
+
+			c.state.clear()
+
+			select {
+			case <-time.After(c.delayFunc(c.reconn)):
+				// If connection drops, we assume that any in-progress tasks
+				// have failed on the server and we have to redo them all.
+				c.batch.clear()
+				c.wip.all(func(t *Task) { c.batch.add(t.value()) })
```
tsmith023: Is there a deadlock risk here between `c.mu.Lock` in `c.wip.all` and `b.mu.Lock` in `c.batch.add`?

Perhaps not but it might be worth running some of the unit tests you have or plan to introduce with the `-race` flag to be sure

internal/api/ssb/protocol.go
```diff
@@ -0,0 +1,656 @@
...
+
+	// Close the batch stream.
+	Close() error
+}
+
+// BatchRequest is a sized container, which accumulates batch items
+// until their size reaches the parent stream's capacity.
+type BatchRequest interface {
+	Add(v any) (added, full bool)
+}
```
tsmith023, line 127: These interfaces appear to only be used internally by the ssb package so should we make them private interfaces?

internal/api/ssb/protocol.go
```diff
@@ -0,0 +1,656 @@
...
+		}
+	}
+}
+
+// size returns the number of entries in the cache.
+func (c *cache) size() int {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return len(c.m)
+}
```
tsmith023, line 660: Worth it to use an `RWLock` here? Probably not tbh!

internal/api/ssb/protocol.go
bevzzz: > Is there a deadlock risk here between c.mu.Lock in c.wip.all and b.mu.Lock in c.batch.add?

I don't think they can deadlock, `c.mu` and `b.mu` are two separate locks and wip cannot access `b.mu` and vice versa.

```
cache.Lock
    for entry in cache {
        batch.Lock
        put value in batch
        batch.Unlock
    }
cache.Unlock
```

> might be worth running some of the unit tests you have or plan to introduce with the -race flag

That's a good idea, I'll add the -race flag to all tests in the pipeline

internal/api/ssb/protocol.go
bevzzz: I think technically yes, _semantically_ it could be an RWMutex, but we need to read-lock the cache rarely enough (for `size` and `all`, which will never happen concurrently) that the basic Mutex is probably good enough. 

I also just realized that `state` has an RWMutex, even though there is max 1 "send" goroutine reading it. So will switch to normal Mutex there too.

P.S.: I found [this thread on Reddit](https://www.reddit.com/r/golang/comments/1fvkb4n/at_what_scale_would_you_ever_choose_to_use_mutex/) pretty informative.

internal/api/ssb/protocol.go
bevzzz: To some extent yes. `NewRequest() BatchRequest` is the exact signature a transport implementation must have, so `BatchRequest` would have to be exported.

Stream can be package-private if we change `Send(BatchRequest)` to `Send(batch any)` (!!), as per the reason above.

internal/api/ssb/protocol.go
bevzzz: I think so. I'd then change `add(v any)` to `add(t *Task)` and read `t.value()` internally in batch, such that `v any` is really confined to `batch` and because then `add` and `addMany` can have similar signatures.

internal/api/ssb/protocol.go
bevzzz: As per it's docs, [CloseSend](https://pkg.go.dev/google.golang.org/grpc#ClientStream.CloseSend) will always return a nil error. Send, OTOH, [may return io.EOF](https://groups.google.com/g/grpc-io/c/XcN4hA9HonI/m/F_UDiejTAwAJ) if the stream is closed, in which case we have to read Recv() until we get the _actual_ error. 

When we do `s.Close()` the adapter will Send(StopStreamRequest) + CloseSend internally. Since EOF is a sentinel error, we can safely discard it.

collections/client.go
bevzzz: Fixed in https://github.com/weaviate/weaviate-go-client/pull/434/commits/0eae5a1b2089a677b15469832e90734d116a0b9b

batch/client.go
bevzzz: > imo, the abstraction is cleaner with private structs that are created using public init methods

But those structs are not private. They are part of the package's public API, defined it terms of types known to the caller.

Similarly to how `api.SearchRequest` has fields `NearText *api.NearText` and `NearVector *api.NearVector`.

> ```go
> ssb.NewObject(o.toAPI())
> ssb.NewReference(r.toAPI())
> ```

By the same token we'd need to introduce `NewStarted() Event`, `NewOOM() Event`, `NewAck() Event` and so on for all server-side events, because Event and Data are both union-like structs (idk what the correct terminology for this is).

Since these are all part of an internal API where we control the code and its usage, I think it will only add boilerplate. This is in contrast to `Distance(d float64) Similarity` in the `query` package, where we use this to steer the user away from misusing the API.

data/client.go
bevzzz: >  and avoid exposing api.BatchObject publicly.

I'm with you on that. We use `ToAPI` in the `batch` package. Perhaps I could move `batch` to `data/batch` and then defined these helpers in `data/internal`. 

This makes sense to me 👍

internal/api/ssb/protocol.go
bevzzz: I like the shorted RetryFunc, but I agree that it's not obviously a predicate.
Renamed to `CanRetryFunc`, matching the `Client.canRetry` field name.

data/client.go
bevzzz: >  Perhaps I could move batch to data/batch and then defined these helpers in data/internal

Tried that and realized it will also require us to define Object, Reference, ObjectPath, and other data types in `data/internal`, then re-export them. We agreed this was too much hassle and that `ToAPI` methods can stay.

internal/api/ssb/protocol.go
bevzzz: Similarly to `Prepare(any)`, `Send()` shouldn't really be asking for `BatchRequest`, because it doesn't care about it's `Add()` method; it's only a guardrail, but a one just as superficial as `Prepare(interface { isBatchData() }`.

There's only 1 place in `protocol.go` where we call Send() so there's not much that can go wrong there. I'm happy to do the changes I detailed in the prev comment.

internal/api/ssb/protocol.go
tsmith023: I like that change to `add(t *Task)` so that `t.value()` is confined to `batch`

internal/api/ssb/protocol.go
bevzzz: Added in https://github.com/weaviate/weaviate-go-client/pull/438

internal/api/ssb/protocol.go
```diff
@@ -0,0 +1,656 @@
...
+	close(s.changed)
+	s.changed = make(chan struct{})
+}
+
+// await blocks until flag is set or ctx expires
+// and returns [Context.Err] in the latter case.
+func (s *state) await(ctx context.Context, af actionFlags) error {
+	// TODO(dyma): test+fuzz
+	s.mu.Lock()
+	for s.flags&af != af {
```
asdine: It's a bit hard to follow, a comment here would be great.

Do you think this can be simplified a bit? To my understanding, an actionFlag can have only two states, not many, right? If so, there's no need for flags, just a byte enum. (I might have misunderstood though)

batch/client.go
```diff
@@ -0,0 +1,97 @@
...
+	"context"
+
+	"github.com/weaviate/weaviate-go-client/v6/data"
+	"github.com/weaviate/weaviate-go-client/v6/internal/api"
+	"github.com/weaviate/weaviate-go-client/v6/internal/api/ssb"
+	"github.com/weaviate/weaviate-go-client/v6/internal/api/transport/stream"
+	"github.com/weaviate/weaviate-go-client/v6/internal/dev"
+)
+
+type Task struct{ *ssb.Task }
```
asdine: nit: Embedding internal types into public ones makes it hard to document them. I'm not sure but it might be hard to read them in the go doc

batch/client.go
```diff
@@ -0,0 +1,97 @@
...
+
+func (t *Task) Wait() error {
+	<-t.Done()
```
asdine: If it's a public method, it would be nice if we could pass a context to Wait to make it cancelable

internal/api/ssb/protocol.go
```diff
@@ -0,0 +1,656 @@
...
+	for {
+		select {
+		case t, ok := <-c.queue:
+			if !ok {
+				goto Drain
+			}
+
+			v, err := c.transport.Prepare(t.data)
+			if err != nil {
+				t.complete(err)
```
asdine, line 362: Just to be sure: no `continue` needed here right?

internal/api/ssb/protocol.go
```diff
@@ -0,0 +1,656 @@
...
+
+		// Start a stream, unblock the 'send' goroutine, and continue reconnecting
+		// until batch completes, the server is deemed unresponsive, or context expires.
+		for ; c.reconn < c.reconnLimit; c.reconn++ {
+
+			var s Stream
+			if s, err = c.transport.NewStream(c.ctx); err == nil {
+				ctx, cancel := context.WithCancel(c.ctx)
+				tick <- span{ctx: ctx, stream: s}
+				if err = c.recv(s, cancel); err == io.EOF {
```
asdine, line 305: We are ignoring other errors, is it because we can only get `io.EOF` and "server blablabla"? If so we should add a small comment.

internal/api/ssb/protocol.go
bevzzz: Great catch! Very much in need of `continue`

internal/api/ssb/protocol.go
bevzzz: Correct: `io.EOF` means the server-side of the stream was closed successfully, and everything else means "reconnect".

Will comment.

batch/client.go
bevzzz: The Client guarantees that all tasks will be finished by the time `Close()` returns, so Wait should not block at all. I added `Wait()` as this really simple one-line helper to be used in exactly this scenario.

The user needs to care about cancelation only if they do it within the .Add loop, which is less common and in which case I'd rather have them write `select { <-t.Done(), <-ctx.Done() }` by hand so they're more careful.

Let me know if you think it's good DX or not.

internal/api/ssb/protocol.go
bevzzz: > an actionFlag can have only two states

The `actionFlags` is rather a set of "permissions". E.g. initially the client can only "preprare" the batch but after we get "Started" event, it can both "prepare" and "send".

```go
case event.Started:
    c.state.set(canPrepare | canSend)
```

Logically, `canSend` also includes `canPrepare`, so I could probably replace the bitmap with enum and just export `awaitCanSend(ctx)` and `awaitCanPrepare(ctx)` or sth to that extent. But it seems like it makes that logical thing implicit.

WDYT?

bevzzz: I will add a comment regardless. Maybe I should rename `actionFlags` to `permissionFlags`?

asdine: No it makes sense if it contains multiple permissions, not just one (I missed the `canPrepare | canSend` part). It makes sense to leave it as it is

batch/client.go
asdine: Are there scenarios where a user would want to wait for a task with a timeout but doesn't want the client to be closed?

```go
router.GET("/foo", func(w, r) {
  // do something with SSB
  // then wait
  err := t.Wait(r.Context())
  if errors.Is(context.Canceled, err) {
     // 
  }
})
```

bevzzz: I should think not. 

It's not clear if a `context.Canceled` means the task has failed or not. A timeout on `Wait(ctx)` can mean:
- client is still waiting for the batch to fill up -> ok
- server is taking a long time to report result for `t` -> maybe ok
- client is trying to reconnect with 30s backoff -> maybe ok
- ???

Given this uncertainty, I just don't think Wait(ctx) is meaningful.

----

OTOH, `c.Add(ctx)` might prove very useful! I used to think of the SSB stream as being relatively short-lived, so Add would use the same context as Client's send/recv routines.
However, it is perfectly reasonable to say: try to add X to the batch but bail out after 30 seconds if the stream is running as a "background" process of a server like in your example (I assume).

batch/client.go
bevzzz: Updated to store `{t *ssb.Task}`. I decided not to export Done and Err for now, see my comment about Wait below.

internal/api/ssb/protocol.go
tsmith023: Hmm, I see the complexity now. Let's leave them public to not overcomplicate things!

batch/client.go
```diff
@@ -0,0 +1,110 @@
...
+type (
+	Config ssb.ClientConfig
+	Option func(*ssb.ClientConfig)
+)
+
+func WithRetryFunc(f func(string, int, error) bool) Option {
+	return func(c *ssb.ClientConfig) {
+		c.CanRetry = ssb.CanRetryFunc(f)
+	}
+}
```
tsmith023: Should this also be called `WithCanRetryFunc`?

internal/api/ssb/protocol.go
```diff
@@ -0,0 +1,669 @@
...
+		state:       newState(canPrepare),
+		canRetry:    conf.CanRetry,
+		delayFunc:   conf.Reconnect.DelayFunc,
+		reconnLimit: conf.Reconnect.Limit,
+	}
+	c.init()
+	return c
+}
+
+// Add puts creates a new tasks and puts it on the work queue.
```
tsmith023: typos 😁

client.go
```diff
@@ -349,15 +354,25 @@ func (et *errorTransport) Do(ctx context.Context, req, dest any) error {
 	return err
 }
 
-func (et *errorTransport) Close() error {
-	if cl, ok := et.t.(io.Closer); ok {
+func (ht *hybridTransport) NewStream(ctx context.Context) (transport.BatchStream, error) {
+	dev.AssertType[stream.Transport](ht.transport, "transport")
+	return ht.transport.(stream.Transport).NewStream(ctx)
+}
```
tsmith023: This type assertion on `ht.transport` to be of type `stream.Transport` remains quite confusing IMO

What is the value of obscuring the requirement that `internal.Transport` is also an implementer of `stream.Transport` in this way? Why not make it more explicit by having `internal.Transport` compose `stream.Transport`? e.g.
```go
package internal
type Transport interface {
    stream.Transport
    Do(...) error
}
```

I wonder whether this overall design has been arrived at due to the departure of the codebase from the "accept interfaces return structs" proverb wrt `transport.newTransport`, which returns `internal.Transport` instead of `transport.transport` thereby avoiding duck-typed interface behaviour and instead depending on type assertions. wdyt?

bevzzz: > the requirement that internal.Transport is also an implementer of stream.Transport in this way? Why not make it more explicit by having internal.Transport compose stream.Transport

It would be more correct to say: the _implementation of internal.Transport used here_ must also implement stream.Transport. 

The reason `internal.Transport` does not compose `stream.Transport` is because 19/20 places where the former is used do not need the streaming capabilities and the interface they work with should remain lean. IMO, if one is extending the `cluster` APIs, they needn't have to wonder why the transport in `cluster.Client` has `MaxSize() int` method and what to do with it.

> I wonder whether this overall design has been arrived at due to the departure of the codebase from the "accept interfaces return structs" proverb wrt transport.newTransport, which returns internal.Transport 

You're quite right there. I intentionally made the signature such that in `client_test.go` the transport could be mocked away. But it does send ripples!
I will change the signature so that `transport.NewFunc` returns the composed interface, in which case we won't need the type assertion at all.

tsmith023 (approved):

PR #437, opened 2026-07-27
v6: Example for SSB usage

bevzzz:


example/batch/main.go
```diff
@@ -0,0 +1,77 @@
...
+	for _, t := range tasks {
+		if err := t.Wait(); err != nil {
+			log.Printf("\t- %s failed: %q", t.ID(), err)
+		}
+	}
+
+	n, err := h.Count(ctx)
+	example.Catch(err)
+
+	log.Printf("Collection %q has %d objects", h.CollectionName(), n)
```
tsmith023: Should this assert that the number of objects is as expected of the test?

bevzzz: I aim to refrain from assertions in these tests, because half of the success depends on the server's performance, which can often be flaky even for things like object count in the collection. 

My thinking was that these examples would serve as smoke tests of a sort, ensuring that _the server understands the requests sent by the client_, i.e. they are valid requests.

tsmith023: My thinking was that these examples should be runnable and workable end-to-end so if they ever start returning unexpected results then this would harm their usability for people coming across them, i.e. the user might start seeing unexpected behaviour they'd need to interpret themselves. wdyt?

bevzzz: > if they ever start returning unexpected results then this would harm their usability for people coming across them

But the usability for people is seeing an example of the API usage in a real codebase, right? Not from the server responding back with a particular object count (in this example).

Take the snippets in [our docs](https://docs.weaviate.io/weaviate/connections/connect-local#no-authentication-enabled) as example: while they are also `runnable and workable end-to-end`, they never assert the results and only print them.

> so if they ever start returning unexpected results

I'd go as far as argue that with SSB you can't really have the "expected" result, because there are many factors that can thwart the expectation: the test server runs out of memory, bad network, an object cannot be inserted, etc.

tsmith023: >But the usability for people is seeing an example of the API usage in a real codebase, right?

Yes but I'm sure users would copy-and-paste example code. Having said that, it's much more likely their agents will do this for them instead. In this case, do we want to be adding the cancellation nuance to this example in case that would poison downstream users' agents when looking at how to use SSB with the Go client. wdyt?

> I'd go as far as argue that with SSB you can't really have the "expected" result

This is false, imo. The expectation of SSB is that all objects added to the batch end up in the server as objects or returned as errors, it's exceptional if they don't. Hence, asserting the number of inserted objects and number of returned errors is beneficial to ensure the protocol is functioning correctly e2e. wdyt?

bevzzz: > do we want to be adding the cancellation nuance to this example in case that would poison downstream users' agents when looking at how to use SSB with the Go client

I can remove the cancellation aspect, no problem

>  all objects added to the batch end up in the server as objects or returned as errors

Agreed

tsmith023 (approved):

PR #441, opened 2026-08-03
v6: Fuzz SSB Client

bevzzz:
This PR adds a fuzzer for SSB client. The idea is to create a simulation environment which closely resembles the real world (connection errors, OOM, shutdowns, etc), but is driven by pseudo-random choices and also runs much much faster.

While we cannot perfectly control the entire process, we can verify that some properties hold. For example:
- the Client doesn't crash or hang
- the Client does not exceed it's reconnect limit
- a batch never contains duplicated items
- eventually most of the data makes it to the server

Owning to the fact that server is essentially the same process, this approach helped uncover a number of less-than-trivial bugs, which this PR also fixes. You can read more about each one under the respective `fix(ssb):` commit, but here's a short rundown:

- Clear `full` flag with `&^=` instead of toggling it with `^=`
- Discard all "retry" items before reconnecting to the stream
- Properly await for 'send' routine to exit before reconnecting
- Atomically consume `canSend` permission when sending the next batch
- Stop `oomTimer` if stream disconnects abruptly
- Close `c.retry` at the end of 'recv' loop, not in `Close()` (race detector)

### DIFF
--- a/internal/api/ssb/protocol.go
+++ b/internal/api/ssb/protocol.go
@@ -166,8 +166,8 @@ type OOM struct {
 }
 
 type Results struct {
-	OK     []string          // IDs of tasks that completed successfully.
-	Failed map[string]string // Batch insertion errors keyed by task ID.
+	OK     []string         // IDs of tasks that completed successfully.
+	Failed map[string]error // Batch insertion errors keyed by task ID.
 }
 
 type permissionFlags uint8
@@ -438,7 +438,7 @@ func (c *Client) recv(s Stream, cancelSend context.CancelFunc) error {
 			retry := make([]*Task, 0, len(failed))
 			if len(failed) > 0 {
 				c.wip.walk(maps.Keys(failed), func(t *Task) (remove bool) {
-					err := errors.New(failed[t.ID()])
+					err := failed[t.ID()]
 					if c.canRetry.check(t, err) {
 						t.retry(err)
 						retry = append(retry, t)
--- a/internal/api/ssb/protocol.go
+++ b/internal/api/ssb/protocol.go
@@ -294,6 +294,7 @@ func (c *Client) init() {
 		defer func() {
 			c.finish(err)
 			close(tick)
+			close(c.retry)
 		}()
 
 		// Start a stream, unblock the 'send' goroutine, and continue reconnecting
@@ -495,8 +496,6 @@ func (c *Client) recv(s Stream, cancelSend context.CancelFunc) error {
 // all accepted tasks have been processed. If the context expires, Close
 // fails all pending tasks and returns the cause of the context's expiry.
 func (c *Client) Close() error {
-	defer close(c.retry)
-
 	close(c.queue)
 	<-c.ctx.Done()
 	err := context.Cause(c.ctx)
--- a/internal/api/ssb/protocol.go
+++ b/internal/api/ssb/protocol.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"iter"
+	"log"
 	"maps"
 	"slices"
 	"sync"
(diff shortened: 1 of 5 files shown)


internal/api/ssb/protocol.go
```diff
@@ -108,9 +108,9 @@ type Transport interface {
 var ErrTooLarge = errors.New("batch item exceeds maximum request size")
 
 type Stream interface {
-	// Send marshaled batch request. The batch is guaranteed to be
-	// of the same type as returned by [Transport.NewRequest].
-	Send(BatchRequest) error
+	// Send marshaled batch request. The batch must be
+	// the value returned by [Transport.NewRequest].
+	Send(any) error
```
bevzzz, line 113: This is motivated by the fact that the `Stream` implementation doesn't need `batch` to implement any interfaces at all, but rather expects it to be the exact same object that `Transport.NewRequest()` returns. Similarly to `Prepare(any)`, `any` is a value that's _opaque_ to ssb.Client.

See [the commit that changed this](https://github.com/weaviate/weaviate-go-client/pull/441/changes/c50bbde8f942265cbfc6b17021876f6d41195e9c)

internal/api/ssb/protocol.go
```diff
@@ -296,30 +298,32 @@ func (c *Client) init() {
...
 				}
-				<-ctx.Done()
 			}
 
 			c.state.clear()
+			// If connection drops, we assume that any in-progress tasks
+			// have failed on the server and we have to redo them all.
+			c.batch.clear()
+			c.wip.all(func(t *Task) { c.batch.add(t) })
+			c.retry = make(chan []*Task, retryBuffer)
```
bevzzz, line 324: No locking is required when writing to `c.retry`, because we can guarantee that no other goroutine accesses it at the moment.

internal/api/ssb/protocol.go
```diff
@@ -231,6 +229,7 @@ func (s *state) await(ctx context.Context, permissions permissionFlags) error {
 			return ctx.Err()
 		}
 	}
+	s.permissions &^= permissions
```
tsmith023, line 232: phwoar, what does this assignment operator do?

internal/api/ssb/protocol.go
```diff
@@ -292,34 +294,37 @@ func (c *Client) init() {
...
 			c.finish(err)
 			close(tick)
+			close(c.retry)
 		}()
 
 		// Start a stream, unblock the 'send' goroutine, and continue reconnecting
 		// until batch completes, the server is deemed unresponsive, or context expires.
-		for ; c.reconnCount < c.reconnLimit; c.reconnCount++ {
-
+		for err = io.EOF; c.reconnCount < c.reconnLimit; c.reconnCount++ {
```
tsmith023, line 302: I'm sure it's necessary but, for my understanding, why is the `err = io.EOF` needed here if we're doing `if err == io.EOF { return }` in the body?

tsmith023 (approved):

internal/api/ssb/protocol.go
bevzzz: So! While `^` (XOR) _flips) the bit, `&^` (AND NOT) _unsets_ it 😮 
I've put together [this Playground example](https://go.dev/play/p/L8YeOUjTykZ) to illustrate that.

I discovered this because previously `batch.clear` would flip! the `full` flag, instead of removing it, so we'd end up with an empty batch which thought it was full.

internal/api/ssb/protocol.go
bevzzz: This was to prevent `c.finish(nil)` from confusing `client.Close` into thinking the context was canceled. Though, to be honest, I can't think of a scenario where we can exit with a nil error (as you correctly pointed out).

I'm inclined to keep this a precaution in case there is actually a codepath which leaves `err == nil` and as a way to say "by default, we should exit with io.EOF". WDYT?

P.S.: Looking at https://github.com/weaviate/weaviate-go-client/pull/441/changes/8bf22dc22704c2cc9ca8dce1bbb6b9796c58d188, I think this addressed a bug at the earlier stage of the testing, probably when the fuzzer was not yet quite stable.

internal/api/ssb/protocol.go
tsmith023: So what exactly does `s.permissions` end up as? `nil`?

bevzzz: No, it just looses the `canSend` bit. Depending on the previous state, either of these 4:

- `(0)`
- `(canPrepare)`
- `(canSend)`
- `(canPrepare | canSend)`

it might turn into `(0)` or `(canPrepare)`. To us it's only important that `canSend` flag is not set, so that the client does not send the next batch until `ACK` arrives and adds `canSend` to permissions.

PR #444, opened 2026-08-05
docs: write the v6 README

g-despot:
The `v6` README is still the v5 one: it points at `go get .../v5` and its only example uses the removed `client.NewClient(config)` / `.Do()` shape.

This replaces it with a pre-release note, requirements, install, a runnable quickstart, the connection options, a short "what changed in v6", and links to the `example/` programs.

Also fixes the build badge, which had a doubly-nested workflow path and was pinned to `?branch=main`.

The Go snippets compile against this branch and against the published `v6.0.0-beta.1` module.


bevzzz:
Is this ready for review yet?

PR #446, opened 2026-08-20
v6: Keepalive configuration

bevzzz:
This PR provides users with a way to configure keepalive for client's underlying transports.

While REST client implicitly uses `http.DefaultTransport`'s 30s keepalive, gRPC has it disabled. In cloud deployments load balancers may therefore drop idle gRPC connections.

We provide a minimal set of parameters to configure the keepalive for both REST and gRPC transports:
- idle time (how soon after the last received byte to ping the server)
- interval (how often to send successive pings if there's no response)
- retry (how many pings can fail before the connection is deemed dead)

In `google.golang.org/grpc` the last two parameters are replaced with [Timeout](https://github.com/grpc/proposal/blob/master/A8-client-side-keepalive.md#basic-keepalive):

> Interval and retry don't quite apply to PING because the transport is reliable, so they will be replaced with timeout (equivalent to interval * retry), the time between sending a PING and not receiving any bytes to declare the connection dead.

and this is exactly what we do to adapt the configuration to gRPC.

See also: https://github.com/weaviate/weaviate-go-client/pull/367

### DIFF
--- a/internal/api/transport/transport.go
+++ b/internal/api/transport/transport.go
@@ -383,6 +383,9 @@ type KeepAlive struct {
 	Retry int
 }
 
+// If k is not nil, dialOptions returns non-nil configuration for
+// [transports.RESTConfig.KeepAlive] and [transports.GRPCConfig.KeepAlive] respectively.
+// If k is nil, both values are nil too.
 func (k *KeepAlive) dialOptions() (*net.KeepAliveConfig, *keepalive.ClientParameters) {
 	if k == nil {
 		return nil, nil
--- a/internal/api/transport/transport.go
+++ b/internal/api/transport/transport.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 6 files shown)


internal/api/transport/transport.go
```diff
@@ -345,3 +350,54 @@ func (t *transport) NewStream(ctx context.Context) (BatchStream, error) {
...
+	// If zero, a default value of 6 is used.
+	//
+	// When applied to [gRPC connections], [keepalive.ClientParameters.Timeout]
+	// is calculated as Interval * Retry.
+	//
+	// [gRPC connections]: https://github.com/grpc/proposal/blob/master/A8-client-side-keepalive.md#basic-keepalive
+	Retry int
+}
+
+func (k *KeepAlive) dialOptions() (*net.KeepAliveConfig, *keepalive.ClientParameters) {
```
tsmith023, line 389: wdyt about a short docstring here making it clear that the first return of the tuple is the HTTP keepalive and the second is the gRPC keepalive? I think it's obvious once you read the code but reflecting it in the comment would make it even easier at a quick glance without needing to read the code

tsmith023 (approved):

PR #453, opened 2026-09-01
fix: expose lastUsedAt, createdAt, apiKeyFirstLetters and namespace on db users

gkampitakis:


internal/api/rbac.go
```diff
@@ -886,6 +901,13 @@ func (ui *UserInfo) UnmarshalJSON(data []byte) error {
 	var resp struct {
 		rest.DBUserInfo
 		UserType string `json:"userType"`
+
+		// The server returns these fields as a nullable type,
+		// e.g. "type": ["string", "null"], which the Swagger -> OpenAPI v3
+		// conversion drops, so they are missing from [rest.DBUserInfo].
+		APIKeyFirstLetters string     `json:"apiKeyFirstLetters,omitempty"`
```
bevzzz, line 908: issue: we also control the conversion code to a certain extent, see `cmd/build.contracts.go`. Seeing that it's a deficiency of the Swagger Converter (external), it feels to me more correct to fix the issue there instead of remedying its effect here.

bevzzz:
Fixed in #454

PR #456, opened 2026-09-03
v6: Boost API

bevzzz:
This PR adds support for Boost API in all query types. Similarly to filters, Boost functions are available in their dedicated package `query/boost`. An intuitive and composable API is the goal here.

```go
c.Query.BM25(ctx, query.BM25{
  Query: "who let the dogs out?",
  Boost: boost.Blend{
    Depth: 4,
    Conds: []boost.Cond{
      { Func: boost.PropertyValue{ Property: "colour" } },
      { Func: boost.Filter{ Filter: filter.Expr{ ... } } },
      {
        Weight: .56,
        Func: boost.TimeDecay{
          Property: "created_at",
          Origin: time.Date(...),
          Offset: new(time.Hour),
          Decay:  new(.23),
        },
      },
      {
        Weight: .11,
        Func: boost.NumericDecay{
          Property: "price",
          Origin: 67,
          Offset: boost.Offset(22),
          Decay:  boost.Decay(88),
        },
      }
    },
  },
})
```

For convenience, all boost functions (`boost.TimeDecay`, `boost.NumbericDecay`, `boost.Filter`, etc) can be used as the `.Boost` parameter directly if only 1 function is applied. See `query/over_all_test.go` for example.


query/boost/boost.go
```diff
@@ -0,0 +1,186 @@
...
+func (td TimeDecay) Expr() api.BoostExpr {
+	return api.BoostExpr{
+		Conds: []api.BoostCond{{Func: td.Func()}},
+	}
+}
+
+type NumericDecay struct {
+	Property string  // Required parameter.
+	Origin   float64 // Required parameter.
+	Scale    float64 // Required parameter.
```
tsmith023, line 102: Should we have validation on the zero values or are they acceptable despite the param being required? (Likewise for the others above)

query/boost/boost.go
```diff
@@ -0,0 +1,186 @@
...
+func (b Blend) Expr() api.BoostExpr {
+	conds := make([]api.BoostCond, len(b.Conds))
+	for i, c := range b.Conds {
+		cond := api.BoostCond{
+			Func:   c.Func.Func(),
+			Weight: c.Weight,
+		}
+		if c.Weight == 0 {
+			cond.Weight = b.Weight
+		}
```
tsmith023, line 51: Do you know whether the server expects this client-side logic to exist or whether it has its own logic for when `cond.Weight==0`?

query/boost/boost.go
bevzzz: I'd leave it to the server, if possible. Property is already [validated server-side](https://github.com/weaviate/weaviate/blob/main/adapters/handlers/grpc/v1/parse_search_request.go#L865). I also don't know that we can validate `origin` because 0 might be a valid value.

query/boost/boost.go
bevzzz: It's something I saw being done in the [Python client](https://github.com/weaviate/weaviate-python-client/blob/e41327362f23572a2ef67ff33ea418d0e66959ad/weaviate/collections/classes/grpc.py#L610-L611), and tbh, simply replicated that here.

I am wondering what happens when `c.Weight == 0`, `b.Weight == .6` and there're, say 3 conditions. Each of them would have a weight of `0.6` and I do not quite know if that's alright.

@trengrj could you help me understand this part ⬆️ please?

tsmith023: I'm happy to merge this to ensure parity, there is likely a reason for it

tsmith023 (approved):

PR #458, opened 2026-09-03
v6: Near Object query

bevzzz:
This PR adds NearObject query.

### DIFF
--- a/query/client.go
+++ b/query/client.go
@@ -23,6 +23,7 @@ func NewClient(t internal.Transport, rd api.RequestDefaults) *Client {
 		defaults:   rd,
 		OverAll:    overAllFunc(t, rd),
 		NearVector: nearVectorFunc(t, rd),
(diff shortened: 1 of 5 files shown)


query/client.go
```diff
@@ -199,9 +201,9 @@ func query(ctx context.Context, t internal.Transport, r request, f func(*api.Sea
 		return nil, nil
 	}
 
-	objects := make([]Object[map[string]any], len(resp.Results))
-	for i, obj := range resp.Results {
-		objects[i] = unmarshalObject(&obj)
+	objects := slices.Grow([]Object[map[string]any](nil), len(resp.Results))
+	for _, obj := range resp.Results {
+		objects = append(objects, unmarshalObject(&obj))
```
bevzzz, line 206: This is to avoid allocating an empty slice.

tsmith023 (approved):

PR #459, opened 2026-09-04
v6: return insert errors as an error value

bevzzz:
Something I felt the need for while migrating server's e2e tests to v6.
This makes error handling for insert that much more ergonomic and idiomatic.

I remember there being a issue in either Go client or the Java client repo (can't find it right now) saying how it's easy to forget to handle partial failure, because checking `err != nil` is not sufficient.

```go
r, err := songs.Data.Insert(Standoff, BlastFurnace)
if err != nil {
  // Handle transport error
}
if len(r.Errors) > 0 {
  // Handle insert errors
}
```

With this refactor the returned err is non-nil if an insert failed for any of the objects in the batch. The user can narrow the error to `data.InsertError` to get the error messages.

```go
r, err := songs.Data.Insert(Standoff, BlashFurnace)
if err != nil {
    var partial data.InsertError
    if errors.As(err, &partial) {
        for id, err := range partial.Errors {
            retryLater(id) // for example
        }
    }
    return err
}
// batch succeeded!
```

### DIFF
--- a/data/client.go
+++ b/data/client.go
@@ -53,8 +53,15 @@ type Object struct {
 }
 
 type InsertResult struct {
-	Took   time.Duration
+	Took time.Duration
+}
+
+type InsertError struct {
 	Errors map[uuid.UUID]string
+}
+
+func (ie InsertError) Error() string {
+	return fmt.Sprintf("insert failed for %d objects", len(ie.Errors))
 }
 
 // Insert new objects into the collection.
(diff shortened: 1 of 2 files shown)


tsmith023 (approved):
I like `errors.As` here but it will need to be specifically documented since it's a marked difference from py/ts where an equivalent of this would be to throw a specific exception that, when caught, contains the internal errors

bevzzz:
> it will need to be specifically documented [...] py/ts where an equivalent of this would be to throw a specific exception

I agree! It feels like it makes a lot more difference in Go, because `err != nil` is so common; in py/ts (Java too) exceptions get treated like "higher level" concerns and only caught at _some_ places in the code.

cc @g-despot

